PR #122, opened 2013-05-11
Validation of submitted parameter keys

dfharmon:
Validation of all passed in parameter keys.  Check each parameter key that is passed in against the api method definition if the validate_keys option is set to true for the config.  If unknown params are passed in an UnknownParam exception will be raised.


iNecas:

PR #124, opened 2013-05-30
A boolean parameters can be configured with both :bool and :boolean

Nerian:
A boolean parameters can be configured with both :bool and :boolean options. A boolean parameter can be any of: 'true', 'false', '1', '0'

### DIFF
--- a/lib/apipie/extractor/collector.rb
+++ b/lib/apipie/extractor/collector.rb
@@ -63,12 +63,12 @@ module Apipie
             # we specify what type it might be. At the end the first type
             # that left is taken as the more general one
             unless param_desc[:type]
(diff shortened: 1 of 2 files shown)


Pajk:
@Nerian @iNecas What do you think about this in 2015:)? I like the change and I don't think it should cause some serious troubles to anyone. Even if it does it will be mentioned in change log for next version so ppl will be aware of it.

Nerian:
Awesome :+1:

iNecas:
Aaaan v0.3.6 with this change has just been released

iNecas:
Thanks for patience @Nerian  :)

Nerian:
wooot!

PR #187, opened 2014-01-23
Instead of manually redefine the route for the action, retrieve it from the routes of the application.

mtparet:
#### Description

Instead of manually redefine the route for the action, retrieve it from the routes of the application.
This enable also to have only place where to maintain the routing.
#### Example:

Instead of defining the name, HTTP verb, and route of the action.

```
api :POST, "/users", "Create user"
```

You could now just define the name of the action.

```
api! 'Create user'
```

HTTP verb and route will be retrieved from the routes of the application.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -193,8 +193,9 @@ api
 api!
   Provide short description and additional option.
   The last parameter is methods short description.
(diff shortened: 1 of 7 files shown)


iNecas:

PR #194, opened 2014-01-28
api keyword without specifying the verb/path loads values from routes

iNecas:


mtparet:
We can close this PR @iNecas

Pajk:
@iNecas Closing this PR as this feature is already in master through @mtparet PR if I understand it correctly:).

PR #209, opened 2014-02-21
Authorization of controllers and methods

candland:
This allows someone to configure custom authorization of controllers and methods.


Pajk:
@candland Thanks for contribution. This could be useful in some cases. I renamed the config option to `authorize` so it's consistent with `authenticate` option. Also did few changes (https://github.com/Apipie/apipie-rails/commit/741bc3d06b1c69fb22a918109d4369796ec6ee1e) but the logic remains the same.

PR #254, opened 2014-05-30
Supply :validations parameter for simple inclusion

jmoses:
This allows the user to outline some purely textual validation text for
inclusion into the api documentation. Useful when clarifying a minor
point of the "type" validator that doesn't warrant a custom validate.

For example, specifying that an +Integer+ parameter must be > 0.

This change also provides more visual weight to the
validator/validations section of the parameter documentation.


Pajk:
@jmoses Nice contribution, thanks!

PR #282, opened 2014-08-26
Add ability to markup validator description

exAspArk:
Hi!
I'm using `config.markup = Apipie::Markup::Markdown.new` and I have my custom validator:

``` ruby
class Apipie::Validator::StringQueryValidator < Apipie::Validator::BaseValidator
  def initialize(param_description, argument, options)
    super(param_description)
  end

  def validate(value)
    !!(value =~ /^[^&*\/\^]+$/)
  end

    if options[:validator] == :query
      self.new(param_description, argument, options)
    end
  end

  def description
    'Must have any characters except: `/` `&` `^` `*`'
  end
end
```

I'd like to make this description highlighted :smiley_cat:

### DIFF
--- a/app/views/apipie/apipies/_params.html.erb
+++ b/app/views/apipie/apipies/_params.html.erb
@@ -17,7 +17,7 @@
       <%= param[:description].html_safe %>
       <% unless param[:validator].blank? %>
         <br>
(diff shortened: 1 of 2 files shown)


iNecas:

PR #285, opened 2014-08-29
Don't specify protocol in Disqus script tag src

chrise86:
Allows Disqus to load in http and secure https environments


iNecas:

PR #286, opened 2014-08-29
Fixes #7284 - BooleanValidator set expected_type as boolean

dustints:


iNecas:

PR #289, opened 2014-09-05
Fix unwanted html encoding.

timscott:
Fix so that raw HTML renders as HTML.


Pajk:
@timscott thanks, merged! https://github.com/Apipie/apipie-rails/commit/dff0d8141003d328921d6dda9a5262b160b798f5

PR #291, opened 2014-09-16
Fixes ConcurrencyError

dwbutler:
We run our app servers in Puma under JRuby. Sometimes an app server gets into a bad state, and starts throwing the following error from Apipie:

```
ConcurrencyError (Detected invalid hash contents due to unsynchronized modifications with concurrent users)
```

We have not been able to consistently reproduce this issue, but it appears that several requests executing on the same controller action at the same time cause the options hash to be destructively modified concurrently. From that point forward, all API calls fail.

We believe the attached patch resolves the issue by avoiding destructive modifications.

### DIFF
--- a/lib/apipie/error_description.rb
+++ b/lib/apipie/error_description.rb
@@ -13,7 +13,7 @@ module Apipie
 
     def initialize(code_or_options, desc=nil, options={})
       if code_or_options.is_a? Hash
-        code_or_options.symbolize_keys!
+        code_or_options = code_or_options.symbolize_keys
         @code = code_or_options[:code]
         @metadata = code_or_options[:meta]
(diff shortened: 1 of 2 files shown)


Pajk:
@iNecas Any reason why not merge this? It seems it's not harmful at all and if it fixes concurrency issues I vote for merging it in.

dwbutler:
FWIW, we've been using this patch in https://www.family.me for the past year and a half without issues. It seems to have fixed the problem for us.

mathieujobin:
please try 0.7.2 or 0.8.1

this should no longer be an issue and moving the symbolize keys does not make much sense to me ... 


please reopen if I am wrong.

PR #300, opened 2014-10-23
Proper url_prefix when doc_base_url has multiple folders

ryanische:
The assignments to `Apipie.url_prefix` assume that the doc_base_url is one folder. This PR corrects assignments to `Apipie.url_prefix` with the proper traversal to the documentation root.


iNecas:

Issue #168, opened 2013-11-01
validators executes after before_filter


freemanoid:
`Apipie` params validators executes _after_ rails `before_filter` so if we want to use some params in `before_filter` we **should validate it manually**.

PR #306, opened 2014-11-15
Allow validations in before filter

dprice-fiksu:
Addresses issue #168, 'validators executes after before_filter'. Adds the ability to call APIpie validations as a before_filter, instead of by wrapping the controller method (but preserves the old behaviour as the default).


iNecas:

PR #310, opened 2014-11-30
Support multipart post

bradrf:
Provides nice bodies when recording tests that POST multipart forms. Will only provide the multipart headers and string bodies, skipping any file contents (instead providing a marker in the output to indicate where the body would have been placed).


iNecas:

PR #318, opened 2014-12-28
Make rubocop happy

burnettk:
adds couple tweaks that make rubocop more happy and should be fine, i believe, since i see you're not supporting ruby 1.8.


iNecas:

PR #327, opened 2015-02-05
The Examples formatting was using string concatenation, I needed more control

mark-ellul:
The examples were being formatted using string concatenation. I needed to have access at the template level.
So I added a new configuration property, render_examples_with_template,  which allows the user to render the examples in the method description from a partial, once set, the new example partial will be used, or that partial can be overridden like the others templates.

NOTE: I couldn't get the Example partial to look like the string concatenation, so thats why its configurable.

### DIFF
--- a/app/helpers/apipie_helper.rb
+++ b/app/helpers/apipie_helper.rb
@@ -6,4 +6,13 @@ module ApipieHelper
     end
   end
 
+  def format_example_data(data)
+    case data
+    when Array, Hash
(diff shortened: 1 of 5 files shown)


iNecas:
Thanks. Looks promissing. What exactly were the issues for not being able to have a template to achieve the same results as with the string concatenation?

mark-ellul:
The Issue was that the string concatenation uses markup that I didn't know how it translates exactly to ERB, and it depends on which markup generator was being used.

PR #329, opened 2015-02-10
Load controllers if not eager loaded

mtparet:
### Description
If not eager loaded, we need to load controllers otherwise no documentation will be generated/displayed.
The logic of reloaded (when already loaded) is moved to the `reload_documentation` method.


iNecas:
Looks good, merging now.

iNecas:
apipie-rails 0.3.2 with this fix was just release. Thanks @mtparet  for the contribution!

PR #332, opened 2015-02-12
Add PATCH to the POST, PUT list

akenger:
It would seem that a PATCH request should behave like a PUT and POST in how the parameters get mapped.  I think typically, since you could be patching a large enough data set, the parameters wouldn't go in the URL but would be submitted as part of the request body.  Rails form builders treat PATCH as a POST with a hidden parameter.


iNecas:
Thanks!

iNecas:
apipie-rails 0.3.2 with this fix was just release. Thanks @akenger for the contribution!

PR #333, opened 2015-02-15
Recursive globbing for controllers

baweaver:
Added a recursive glob to the default configuration to prevent the need to specify these things, and put the base url in for a more accurate result.


iNecas:
Does this really work. AFAIK the @api_base_url is a hash (with versions as keys). But +1 for the globbing

baweaver:
Yeah, granted. I'll patch that real quick. I assumed it was a string.

baweaver:
Done, that should be safer.

iNecas:
Thanks @baweaver. Merging now

iNecas:
apipie-rails 0.3.2 with this fix was just release. Thanks @baweaver for the contribution!

PR #334, opened 2015-02-18
Update README.rst

baweaver:
Added a caveats section to testing, as this is a common issue that will likely come up fairly frequently.


iNecas:
Thank you @baweaver

PR #339, opened 2015-03-12
Readme fixes

emma-borhanian:
* `app_info` updated to match https://github.com/Apipie/apipie-rails/blob/4627d69fd64766d4f973f3603599d9d0a53d8046/spec/dummy/config/initializers/apipie.rb#L51

```ruby
  # set default version info, to describe specific version use
  # config.app_info[version] = description
  # or put this in your base or application controller
  config.app_info = "Dummy app for testing"
```

* `RailsFormatter` has been renamed to `RoutesFormatter`


iNecas:
Thanks @emma-borhanian!

PR #342, opened 2015-03-20
Including brasilian portuguese internationalization

dadario:
Just including internationalization for brasilian portuguese.


iNecas:
Nice, could I ask you to squash the commits to one?

iNecas:
Also, #341 addes some more strings to translate, you might want to incorporate those as well

PR #348, opened 2015-03-31
added heading helper_method to controller for when helper is not loaded

saneshark:
Fixes #251

Issue #251, opened 2014-05-24
Rails 3.0.x: undefined method *heading*, issue with ActionDispatch::Reloader


jensb:
Using Rails 3.0.20 and Ruby 1.9.3, I get this error when trying to access the apipie index:

*NoMethodError in Apipie/apipies#index*

Showing .../ruby1.9/gems/1.9.1/gems/apipie-rails-0.1.3/app/views/apipie/apipies/_method_detail.erb where line #34 raised:

*undefined method `heading' for #<#<Class:0x007fabe6e27770>:0x007fabe6ddc658>*
Extracted source (around line #34):

    31: <% end %>
    32: 
    33: <% unless method[:params].blank? %>
    34:   <%= heading('Params', h_level) %>
    35:   <table class='table'>
    36:     <thead>
    37:       <tr>

If this is Rails-3.0 specific: I know that Rails 3.0 is deprecated, but if a fix is trivial (it's in a template, after all), can you provide a patch? 
Thank you!

saneshark:
:bug: This is an issue on:  `rails (4.0.12)` and `ruby 2.1.5`

saneshark:
This fixed the issue for my project, but I'm wondering if maybe it is a bad idea to have heading in a Helper altogether, as the same method name could be implemented for a given project and would simply override the method defined here as the method call is not namespaced from what I can tell. 

It would be better to get rid of `ApipieHelper` altogether and add it in to the controller with 

```ruby
  def heading(title, level=1)
    content_tag("h#{level}") do
      title
    end
  end
  helper_method :heading
```

This way the method defined within the project could be overridden at the controller level. Attempting this resulted in 5 failed tests however.

saneshark:
@iNecas this seems to have resolved the issue for my project.

PR #350, opened 2015-04-06
fix undefined method `content_tag`

emma-borhanian:
Started getting the following exception after bumping from 0.3.3 to 0.3.4:

```
ActionView::Template::Error:
 undefined method `content_tag' for #<Apipie::ApipiesController:0x007ffdb1a5e608>
# /Users/emma/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/apipie-rails-0.3.4/app/helpers/apipie_helper.rb:4:in `heading'
# /Users/emma/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/actionpack-4.2.1/lib/abstract_controller/helpers.rb:67:in `heading'
# /Users/emma/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/apipie-rails-0.3.4/app/views/apipie/apipies/_method_detail.erb:27:in `___sers_emma__rbenv_versions_______lib_ruby_gems_______gems_apipie_rails_______app_views_apipie_apipies__method_detail_erb__4191062763070027136_70363792007840'
# /Users/emma/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/actionview-4.2.1/lib/action_view/template.rb:145:in `block in render'
```

### DIFF
--- a/app/helpers/apipie_helper.rb
+++ b/app/helpers/apipie_helper.rb
@@ -1,4 +1,5 @@
 module ApipieHelper
+  include ActionView::Helpers::TagHelper
 
   def heading(title, level=1)
     content_tag("h#{level}") do


IlyaBylich:
:+1: for this fix, I have same issue on rails 3.2.11

saneshark:
I think this issue probably started manifesting itself for you guys after my PR #348 which addressed #251 was merged. The helper method was not being properly included as part of `ActionController::Base` but now I'm wondering how it is that that this worked for RailsApi prior to my patch?

It might also be worth testing that an existing helper having the same name `heading` within a projects ApplicationHelper doesn't override this method and have unintended consequences. To my knowledge these would all be in the same helper method namespace. Ideally the method would just be available in the context of this specific controller as a method and not pollute the global helper namespace.

lumir:
I have the same issue, I can't use any of my def_param_group in my controller

iNecas:
Thanks

adityashankert:
hey we are trying to use the gem and someone have deleted the branch could someone do this pull to master or should i raise a pull request?

PR #352, opened 2015-04-07
Added russian locale

Le6ow5k1:


config/locales/ru.yml
```diff
@@ -0,0 +1,29 @@
+ru:
+    apipie:
+        resources: Ресурсы
+        resource: Ресурс
+        description: Описание
+        no_docs_found: Документация не найдена
+        no_docs_found_descr: Мы не нашли никакой документации для вашего АПИ.
```
IlyaBylich: Probably it's better to change `АПИ` to `API`

config/locales/ru.yml
```diff
@@ -0,0 +1,29 @@
...
+        follow_instructions_html: Перейдите по ссылке %{href}, чтобы узнать как описать ваше АПИ.
```
IlyaBylich: `ваш API` instead of `ваше АПИ`

IlyaBylich:
In general :+1:, looks good!

config/locales/ru.yml
Le6ow5k1: I think you right, `АПИ` looks ugly

iNecas:
спасибо

Issue #355, opened 2015-04-10
Ability to hide method from documentation


davidcollie:
Simliar to https://github.com/Apipie/apipie-rails/issues/202

There are times we want to hide an entire API method for the same reasons outlined in issue 202.

@iNecas would you be open to a PR in the same style as @nathanhoel s PR https://github.com/Apipie/apipie-rails/pull/208 ?

Example:

```
api :GET, '/users/see_another', 'Boring hidden method that we do not want clients to know about yet'
show false
see 'development#users#create'
see 'development#users#index', "very interesting method reference"
desc 'This method is boring, look at users#create.  It is hidden from documentation.'
def see_another
  render :text => 'This is very similar to create action'
end
```

nathanhoel:
Beat me to it. That was the logical next step for us as we have some internal APIs that we would like to use the validation but not publicize the API docs.

IlyaBylich:
@davidcollie Probably is should be renamed to `show_in_doc false`.
But definitely it's a good idea. :+1: for doing that

iNecas:
+1 for `show_in_doc`, but I see that we have :show already as a key in params.

Alternatively, what do you think about having the :show option in `api` keyword?

davidcollie:
I can rename to `show_in_doc` if that's preferred.  I was going for consistency with the previous usage of `show`.

I kind of prefer having `show` or `show_in_doc` separately.  It really calls it out that the documentation will be hidden.  I can see it being missed when quickly scanning code.

jeremywadsack:
Isn't this what `config.ignored` does?

davidcollie:
`config.ignored` completely skips over the method or controller.  This just hides the docs.  Things like parameter validation will still happen.

Pajk:
Closing this one as the feature is merged and it seems there were no issues with the `show` name since april.

AzRu:
So since `config.ignored` skips over the method/controller and doesn't do any validation, is there a way to hide endpoints in the docs while retaining validation? Using `:show false` at the `api` keyword doesn't work

PR #356, opened 2015-04-10
Hide controller methods from documentation

davidcollie:
Reasons for feature and a description of the functionality described in issue https://github.com/Apipie/apipie-rails/issues/355

Example:

```
api :GET, '/users/see_another', 'Boring hidden method that we do not want clients to know about yet'
show false
see 'development#users#create'
see 'development#users#index', "very interesting method reference"
desc 'This method is boring, look at users#create.  It is hidden from documentation.'
def see_another
  render :text => 'This is very similar to create action'
end
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -224,6 +224,9 @@ see
 meta
   Hash or array with custom metadata.
 
+show
+  Resource is hidden from documentation when set to false (true by default)
+
 Example:
 ~~~~~~~~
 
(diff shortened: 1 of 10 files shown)


davidcollie:
I had to merge in PR https://github.com/Apipie/apipie-rails/pull/350 for testing purposes.

I can refactor the DSL name from `show` to something else if there is a strong objection to that name.

iNecas:
It seems nobody objects on the name. Mergning now. Thanks @davidcollie

PR #370, opened 2015-06-10
Change bindings project to apipie-bindings

domcleal:
Change old reference to foreman_api to apipie-bindings.


iNecas:
Good catch

PR #373, opened 2015-06-18
Display engine path in documentation when using api!

neilgupta:
If my_engine is mounted at '/engine', ApiPie documentation will now
include "/engine" as part of the path when using api!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,8 @@ source "https://rubygems.org"
 
 gemspec
 
+gem 'test_engine', path: 'spec/dummy/components/test_engine', group: :test
+
 # load local gemfile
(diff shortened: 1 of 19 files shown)


neilgupta:
Any plans to merge this? Would be nice to get back on official branch :)

Pajk:
@neilgupta Could you provide a working example or preferably include tests for this? Thanks!

neilgupta:
Finally got around to adding specs for the engine mount path change... Unfortunately I also had to add a hack to workaround #415 that checks the Rails version to get the rails42 tests to pass. That clearly needs a better fix, but it falls outside the scope of this PR.

iNecas:
Only 4 years to get this in, but it made it. Merged in https://github.com/Apipie/apipie-rails/pull/673

PR #377, opened 2015-06-24
Add an alternative to action aware required param

mourad-ifeelgoods:
Feature
======
In some use cases, we want a param to required depending of the action. we could use action aware params (see https://github.com/Apipie/apipie-rails#action-aware-params), but how to handle parameters optional on creation but required on update ?

A possible solution is to pass a list of actions for which to parameter is required (See examples below)

**Make a param required for all actions**

```ruby
param :name, String, required true
```

**Make a param required only for update action**
```ruby
param :name, String, required :update
```

**Make a param required only for update and create actions**

```ruby
param :name, String, required [:update, :create ]
```

*Note*
Beware that the required param may still be overridden by action aware param if specified

### DIFF
--- a/lib/apipie/param_description.rb
+++ b/lib/apipie/param_description.rb
@@ -47,11 +47,7 @@ module Apipie
       @parent = @options[:parent]
       @metadata = @options[:meta]
 
-      @required = if @options.has_key? :required
-        @options[:required]
-      else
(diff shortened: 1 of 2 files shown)


mtparet:
:+1:

iNecas:
Nice, thanks! Sorry for the delay, I was a bit late with the PRs lately.

PR #379, opened 2015-06-25
Allow , and . in numbers, so decimals are allowed

saten:
Can't manage to use validation on decimal values supplied via a json api without this. The ',' allows number with locale other than english to be used, too

### DIFF
--- a/lib/apipie/validator.rb
+++ b/lib/apipie/validator.rb
@@ -391,7 +391,7 @@ module Apipie
       end
 
       def self.validate(value)
-        value.to_s =~ /\A(0|[1-9]\d*)\Z$/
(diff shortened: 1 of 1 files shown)


iNecas:
Maybe this is a good candidate to separate validator (either in here or in your code-base), due to compatibility issues: somebody already might depend on this behaviour

muescha:
note: The dot operator `.` does not need to be escaped inside of a character class `[]`

Pajk:
@saten I agree with @iNecas in this matter. I really wouldn't like number validator to accept e.g. 10.000,50 as a valid number because to me it's not a valid number at all. So let's leave this up to users to create their custom validators if something more advanced is needed.

PR #380, opened 2015-07-01
rdoc is needed when in production as well

frenkel:
When deploying this to production I received errors about rdoc. This commit fixes it.


iNecas:
We don't want to include that as runtime dependency, because the the rdoc markup is optional (and this would mean all the users, even those that use some other markup library would end up with installing this). We try to keep the number of runtime dependencies as low as possible (and rubygems don't support notion of optional dependencies). Therefore, what you should do is adding the dependencies to your project, Closing the PR, hoping the provided solution would help you instead.

PR #386, opened 2015-07-02
Update to the latest jQuery 1.x.

voxik:
Please note that I am not exactly sure how to test this. I've tried this with https://github.com/Apipie/apipie-demo and can't see any difference in behavior, so it hopefully works.

BTW I am submitting this PR, since on Fedora, I'd like to replace the bundled version of jquery by its system version.


Pajk:
@voxik Thanks for pointing this out. I updated the jquery in https://github.com/Apipie/apipie-rails/commit/fd6d385d515ec8d121cf7cae1f766da4fbeecd46 but I left the version number in the filename because I prefer to see the version without opening the file.

voxik:
This  [1] is how the unbundling is done on Fedora. I.e. if you keep the version number in the file name, it might happen that it will state that it is 1.11.3 but when Fedora decides to update the jquery package, it might be in reality different version.

Version in file name is never good idea for packaging.

[1] http://pkgs.fedoraproject.org/cgit/rubygem-apipie-rails.git/tree/rubygem-apipie-rails.spec#n65

Pajk:
@voxik I see...I didn't know about this. Makes sense. Another thing is that you would have to change the build script every time we update jquery and that could be easily forgotten. I'll remove the version from the filename then.

voxik:
Thanks. I appreciate that!

Pajk:
Done in 326b211bf5e4304df5a9963e6a03c1a1d9611d47

PR #389, opened 2015-08-17
Make it easier to install apipie-rails

vassilevsky:
The absence of dollar signs will make it possible to copy, paste and execute all three commands at once.


iNecas:
Thanks!

Issue #360, opened 2015-05-21
Option to allow custom required params error messages


holyxiaoxin:
Right now params error messages are based on the params names. Sometimes, param names does not reflect the input labels from client. Any alternative solutions would be great. (:

Pajk:
@holyxiaoxin thanks to @jcalvert it's now possible to set your custom message for missing parameters. Use `gem 'apipie-rails', :github => 'apipie/apipie-rails', :branch => 'master'` in your `Gemfile` or wait until next version is released.

PR #390, opened 2015-08-26
Add custom message support for missing parameters 

jcalvert:
`ParamMissing` can take the entire `ParamDescription` object instead of just the name, allowing it to interrogate the object to produce a custom missing message instead of the boilerplate `"Missing parameter #{@param.name}"`. Support is for a basic string or function. 

A primary use case for this is when an API is returning an error message that a client can then surface directly to a user without having to maintain a mapping of parameter names to human readable names IE 
`Missing parameter user_id` to `A user must be supplied to complete this action.`. An additional benefit in allowing a `Proc` is that we can evaluate runtime conditions to create the message, thereby allowing internationalization and providing greater usefulness in the error message.

This would also address the feature request in issue #360

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -291,6 +291,10 @@ meta
 show
   Parameter is hidden from documentation when set to false (true by default)
 
+missing-message
+  Specify the message to be returned if the parameter is missing as a string or Proc.
+  Defaults to ``Missing parameter #{name}`` if not specified.
+
 Example:
 ~~~~~~~~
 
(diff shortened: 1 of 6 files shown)


Pajk:
@jcalvert Nice! I find that very useful. Could you also mention this in the readme?

jcalvert:
@Pajk Updated branch and added changes to README under the Parameter Description section.

iNecas:
v0.3.6 with this change has just been released. Thanks @jcalvert

PR #393, opened 2015-09-23
Fix the names of interpolation arguments

sun-zheng-an:


Pajk:
@sun-zheng-an thanks, nice catch

Issue #402, opened 2015-11-10
Empty string not recognized by allow_nil


joshpfosi:
A usage issue: Due to the current implementation of `validate`, a parameter with `allow_nil` set will still be rejected if that parameter is the empty string as `"".nil? #=> false`. I propose changing the conditional check in `validate` to:

```ruby
      return true if @allow_nil && value.to_s.blank?
```

This way both `nil` and `""` are acceptable. I intend to make a fork with this change and will generate a PR. I was wondering, however, if this is perhaps a feature not a bug?

Thanks!

Pajk:
Please see https://github.com/Apipie/apipie-rails/pull/403#issuecomment-164144590

PR #403, opened 2015-11-10
Fixed allow_nil bug

joshpfosi:
See #402 for details!

### DIFF
--- a/lib/apipie/param_description.rb
+++ b/lib/apipie/param_description.rb
@@ -70,7 +70,7 @@ module Apipie
     end
 
     def validate(value)
-      return true if @allow_nil && value.nil?
+      return true if @allow_nil && value.to_s.blank?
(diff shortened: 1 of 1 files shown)


iliabylich:
Method `blank?` returns true for:
+ `nil`
+ `false`
+ empty string
+ empty array
+ empty hash

Why do you invoke `to_s` first? Looks like `value.blank?` would be enough.

joshpfosi:
Your totally right I just wasn't aware of that! Fixed.

Pajk:
@joshpfosi I don't see it as a bug. From documentation:

    allow_nil: Setting this to true means that nil can be passed.

If you want to allow empty string and empty collections using `blank?` I suggest you add another option called `allow_blank`. I would be happy to merge that as a new feature but I'm not going to merge this one, sorry.

PR #405, opened 2015-11-25
Update README.rst

type-face:
configure.validate_keys does not work. Should be validate_key.


Pajk:
@type-face good catch

PR #407, opened 2015-12-02
Add Turkish translations

halilim:


Pajk:
@halilim Thanks for contribution! I don't understand a word but seems legit:).

halilim:
@Pajk You're welcome. I mostly tried to follow the "official" rules. But you can try Google Translate to see that it at least doesn't contain profanity :P

iNecas:
v0.3.6 with this change has just been released. Thanks @halilim

PR #409, opened 2015-12-12
Nerian layout fix

Pajk:


iNecas:
v0.3.6 with this change has just been released, thanks @Pajk  :)

PR #412, opened 2015-12-13
Add viewport meta tag.

buren:
Thanks for an awesome gem :+1: 

I just added a viewport meta tag to make the documentation easier to read on mobile devices.

## Before
before

## After
after

Tested on Android chrome (and chrome/safari/firefox on desktop).

Thanks again :)

### DIFF
--- a/app/views/layouts/apipie/apipie.html.erb
+++ b/app/views/layouts/apipie/apipie.html.erb
@@ -3,6 +3,7 @@
 <head>
   <title><%= t('apipie.api_documentation') %></title>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
   <%= Apipie.include_stylesheets %>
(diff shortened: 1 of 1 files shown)


Pajk:
@buren thanks! Tested on tablet and it's definitely more readable. Still not very nice on a phone but I guess it would need more changes to make it really mobile friendly.

buren:
Yeah, I was meaning to ask. Do you have a plan for the design? I'm using apipie in a new project, quite extensively, and will probably look into making the docs more mobile friendly, please let me know if there is anything I can do to help.

Thanks again for an awesome project, its been a breeze setting it up :) :star:

Pajk:
All thanks should go to @iNecas who has been maintaining Apipie for more than 3 years and of course the community is doing great job too.

AFAIK there are no plans for redesign. I think most of the ppl are using their own design but it would be nice to have something better looking as a default.

iNecas:
v0.3.6 with this change has just been released. Thanks @buren

buren:
Thank you @iNecas :)

PR #420, opened 2016-01-23
Replace #before_filter with #before_action

buren:
Fixes

```
DEPRECATION WARNING: before_filter is deprecated and will be removed in Rails 5.1. Use before_action instead. (called from block (2 levels) in <class:Railtie> at /Users/someusername/.gem/ruby/2.3.0/gems/apipie-rails-0.3.5/lib/apipie/extractor.rb:12)
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -528,7 +528,7 @@ default_version
 validate
   Parameters validation is turned off when set to false. When set to
   ``:explicitly``, you must invoke parameter validation yourself by calling
-  controller method ``apipie_validations`` (typically in a before_filter).
+  controller method ``apipie_validations`` (typically in a before_action).
   When set to ``:implicitly`` (or just true), your controller's action
   methods are wrapped with generated methods which call ``apipie_validations``,
   and then call the action method. (``:implicitly`` by default)
@@ -726,17 +726,17 @@ is raised and can be rescued and processed. It contains a description
 of the parameter value expectations. Validations can be turned off
 in the configuration file.
 
-Parameter validation normally happens after before_filters, just before
+Parameter validation normally happens after before_actions, just before
 your controller method is invoked. If you prefer to control when parameter
 validation occurs, set the configuration parameter ``validate`` to ``:explicitly``.
 You must then call the ``apipie_validations`` method yourself, e.g.:
 
 .. code:: ruby
 
-   before_filter: :apipie_validations
+   before_action: :apipie_validations
 
-This is useful if you have before_filters which use parameter values: just add them
-after the ``apipie_validations`` before_filter.
+This is useful if you have before_actions which use parameter values: just add them
+after the ``apipie_validations`` before_action.
 
 TypeValidator
 -------------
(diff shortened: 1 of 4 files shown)


buren:
Hm, the test suite fails on rails 3.2, because `before_action` was introduced in rails 4. I guess one thing could be to alias? 

Relevant code in rails: [actionpack/lib/abstract_controller/callbacks.rb](https://github.com/rails/rails/blob/master/actionpack/lib/abstract_controller/callbacks.rb#L183-L188)

Pajk:
Maybe it's time to drop Rails 3 support? @iNecas

iNecas:
I'm ok with dropping rails 3, provided we bump a version number. Maybe it's time for v1.0.0?

buren:
@iNecas, @Pajk maybe it could be an idea to wait a little bit for Rails 5 and see if there are any things that needs to be addressed. If there is anything it could be fixed and then v1.0 could be stable for a long time. Otherwise there is a risk you would need to release a v2 in order to make the braking changes to support Rails 5. On the other hand I've actually tried running apipie-rails under Rails 5 and everything seems to be working :)

See: https://github.com/justarrived/just_match_api/pull/155
Heroku review app running rails 5:  https://just-match-api-staging-pr-155.herokuapp.com/api_docs

Thanks for an awesome gem @iNecas and @Pajk

buren:
Revising my above comment, I actually think a v1.0.0 might be a good idea @iNecas. Can't really see what Rails 5 would brake, since its already in beta2 and an rc is probably coming any day now. 

Dropping Rails 3 doesn't seem unreasonable at this stage either, since it soon will be 2 versions old when Rails 5 comes out. Those on Rails 3 can probably live with the current version a while longer :)

buren:
Related to Rails 5 and might be something that would be nice to fix before a v1.0.0: #437

teaspoon:
Hey guys, what is the hold up with the above pull request?

buren:
@teaspoon basically its up to @Pajk and @iNecas to decide about:

> I'm ok with dropping rails 3, provided we bump a version number. Maybe it's time for v1.0.0?
> https://github.com/Apipie/apipie-rails/pull/420#issuecomment-183416040

voxik:
#473 is probably duplicate of this, although it suggests more universal solution ...

alepore:
+1 dropping rails 3 and maybe keeping a branch for that ("for rails 3, use the 0-3-stable branch")

seanhandley:
Someone please merge this 👍

ened:
@iNecas do you need help from the community to get it merged?

iNecas:
Just reminding is ok: we are close to get us more compatible with rails 5 (including tests) https://github.com/Apipie/apipie-rails/pull/527, after that is merged, this will just become green and can be next to go, I expect the release by the end of the week

iNecas:
Merged as 08be2409970c1d25b9b0ca86cf224ba522234048, thanks @buren

iNecas:
apipie-rails-0.4.0 has just been released, including this fix. Thanks @buren for help!

PR #423, opened 2016-01-27
Unify indentations across all locale files

springerigor:
Two spaces per identation level are preferred in Ruby world.

It also updates `JavaScript` word in Polish translation.


Pajk:
important for ppl with OCD:)

iNecas:
v0.3.6 with this change has just been released. Thanks @springerigor

PR #425, opened 2016-02-01
Fix layout path (Rails 4.2.5.1 compatibility)

halilim:
Fixes #421 (Missing template ../../layouts/apipie/apipie ... error). Credits: @FooBarWidget's [comment](https://github.com/Apipie/apipie-rails/issues/421#issuecomment-177998315).

PR #426, opened 2016-02-02
Render enum param values as CODE

febeling:
If an resource enum parameter has values that end in whitespace or dot
it can be hard to tell with certainty from the generated
documentation. This is because values aren't quoted or otherwise
enclosed. This patch encloses them in CODE elements, which stand out
with clarity from surrounding text.

Before:
screen shot 2016-02-02 at 21 23 18

After:
screen shot 2016-02-02 at 21 20 26

### DIFF
--- a/lib/apipie/validator.rb
+++ b/lib/apipie/validator.rb
@@ -149,7 +149,8 @@ module Apipie
       end
 
       def description
-        "Must be one of: #{@array.join(', ')}."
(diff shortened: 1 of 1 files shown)


febeling:
PR #425 fixes the CI failure.

iNecas:
v0.3.6 with this change has just been released, thanks @febeling

febeling:
@iNecas cool, thanks for the headsup!

PR #431, opened 2016-02-12
new validator - DecimalValidator

kiddrew:
Simple patch - allow number params to have decimal places.

### DIFF
--- a/lib/apipie/validator.rb
+++ b/lib/apipie/validator.rb
@@ -374,6 +374,27 @@ module Apipie
       end
     end
 
+    class DecimalValidator < BaseValidator
+
+      def validate(value)
+        self.class.validate(value)
+      end
+
+      def self.build(param_description, argument, options, block)
+        if argument == :decimal
+          self.new(param_description)
+        end
+      end
+
+      def description
+        "Must be a decimal number."
+      end
+
+      def self.validate(value)
+        value.to_s =~ /\A[0-9]+([,.][0-9]+)?\Z$/
+      end
+    end
+
     class NumberValidator < BaseValidator
 
       def validate(value)
--- a/lib/apipie/validator.rb
+++ b/lib/apipie/validator.rb
@@ -391,7 +391,7 @@ module Apipie
       end
 
       def self.validate(value)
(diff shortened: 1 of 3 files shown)


Pajk:
It seems this is quite often requested feature. We wanted to include only few basic validators as it's simple to add custom one. But I'm ok with having a 'decimal' validator included in Apipie. Could you refactor this pull request and create a new 'decimal' validator leaving the 'number' validator untouched? It would be awesome if you mention it also in the readme. Thanks!

kiddrew:
I'd be happy to.  I don't see where the :number validator is even documented in the readme.

kiddrew:
I have the code updated for the validator.  I haven't updated the README yet since it's not clear how I should do that.

lib/apipie/validator.rb
```diff
@@ -374,6 +374,27 @@ def description
...
+          self.new(param_description)
+        end
+      end
+
+      def description
+        "Must be a decimal number."
+      end
+
+      def self.validate(value)
+        value.to_s =~ /\A[0-9]+([,.][0-9]+)?\Z$/
```
Tybot204: What do you all think of validating for negative decimals as well? This would keep this validator more in line with the `IntegerValidator`, which also allows negatives:
```ruby
value.to_s =~ /\A^[-+]?[0-9]+([,.][0-9]+)?\Z$/
```

RocKhalil:
Why this isn't merged ??
Can't find any valid way to use `decimal` without a monkey-patch..

lib/apipie/validator.rb
kiddrew: Yeah, good call

kiddrew:
I just updated to allow for negatives in DecimalValidator. Please merge this in!

iNecas:
Thanks for patience. We are this close `'  '`  to merge this. Would you mind adding a short info to README's validators section about this addition that others can learn easily about it's availability?

kiddrew:
Added. Let me know if that works

iNecas:
Thanks @kiddrew

iNecas:
Apipie-rails-0.5.9 with this change has been pushed to Rubygems.

Issue #436, opened 2016-02-16
Validators is not localized


Arugin:
Strings like "Must be <type>" is hardcoded into validators. It should be moved to en.yml too.

PR #439, opened 2016-02-22
#436 Validators is not localized

clamoris:
#436 Validators is not localized

@iNecas @Arugin I'm not sure if expected types should be localizable. For now, only visible 'Must be ..." are included.


clamoris:
@iNecas ping =)

PR #444, opened 2016-03-17
Add ArrayOf validator

ywen:
### The intended usage

```ruby
param :ids, ArrayOfValidator.for_class, of: Integer
```

### Explanation

Instead of simply checking the value is an Integer, it will try to find
a validator for that type.

### Motivation

Rails of pass Integers as string. For example, instead of `ids: [1, 2]`
it could be `ids: ["1", "2"]`. In this case, we would still think this
is valid input.

But if we do

```ruby
param :ids, Array, of: Integer
```

an error will be raised for `"1"` is not an number

### Miscs

I realize the existence of #189 but that PR doesn't solve this particular problem. Nor did `ArrayClassValidator`

If you fee this PR is worthy I will add README section

Thanks!

### DIFF
--- a/lib/apipie/validator.rb
+++ b/lib/apipie/validator.rb
@@ -247,6 +247,50 @@ module Apipie
       end
     end
 
+    class ArrayOfValidator < Apipie::Validator::BaseValidator
+      class << self
+        def for_class
+          :array_of_validator
+        end
+
+        def build(param_description, argument, options, block)
+          if argument == for_class
+            new(param_description, options, block)
+          end
+        end
+      end
+
+      def initialize(param_description, options, block)
+        super(param_description)
+        @items_type = options[:of]
+        @options = options
+        @block = block
+      end
+
+      def validate(values)
+        values ||= []
+        return false unless values.respond_to?(:each) && !values.is_a?(String)
+        values.all? { |v| validate_item(v) }
+      end
+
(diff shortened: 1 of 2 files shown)


squidget:
When you have an Array input, the main use cases are:
- an array of fixed length, where each index has its own validator.  This is covered by NestedValidator
- an array of variable length, where all elements must be of the same standard type (String, Integer, etc.)  This is covered by Array validator with the of: option.
- an array of variable length, where each element must pass some custom validation (eg. an Integer between 0 and 100).  There is no way to do this currently using the Array or NestedValidator.

Say I have written a custom validator that responds to :numeric to validate numbers that has optional parameters :min and :max to do optional validation of the range.  I could do something like this:

`param :percent, :numeric, :min => 0, :max => 100, :desc => 'Percentage value'`

But what if I had an array, for which all values must be a number between 0 and 100?  The of: parameter doesn't work in this case.

I think this solution goes some way towards covering this scenario, but not completely.  What you'd ideally want is a way to specify another validator for all the elements of the array, including custom parameters to configure that validator

iNecas:
I like the idea of this PR and if I understand @squidget, the suggestion is not to limit the params to just the `:of`, but actually taking the rest of params (perhaps except the ones like :desc) and pass it to the sub-validator. That would make sense as well.

PR #457, opened 2016-05-04
Remove json gem dependency

domcleal:
Ruby 1.9+ ships json in the standard library, so a dependency on the
gem isn't necessary. See also rails/rails@f3433f7c.


domcleal:
Rebased on master.

iNecas:
Thanks @domcleal

PR #459, opened 2016-05-16
Fix json documentation generation.

zjhuntin:
Before it was translating json from html, which could cause characters like ' to be translated to \&#39; instead of being properly translated.

### DIFF
--- a/lib/apipie/param_description.rb
+++ b/lib/apipie/param_description.rb
@@ -206,7 +206,7 @@ module Apipie
     end
 
     def preformat_text(text)
-      concern_subst(Apipie.markup_to_html(text || ''))
+      concern_subst(text || '')
     end
 
     def is_required?


iNecas:
Could you describe specific case that started causing this issue, ideally in a form of a testcase that would be fixed by the patch?

zjhuntin:
Sorry, what I had typed in escaped. Right now documentation generated that contains any ' seems to come out as a ’ if you look at it in json format. We noticed this first in the Katello apipi docs.

zjhuntin:
@iNecas Looking more into this problem, I don't think my way to fix it is correct but I'm not sure what is the proper fix is.

Currently to_json converts the description of an endpoint to html before inserting it into json https://git.io/vrEkA This method changes this into html using the RDoc convert method for it's MarkUp object, https://git.io/vrELU https://git.io/vrEkH This converts it into html with special characters escaped.

So I'm wondering why the text is being converted into html before putting it into json? If we want it make sure that it's html safe, is there another way that we can do that without completely converting it to html?

Issue #466, opened 2016-06-16
allow `error` to support Rails status symbols


alex-tan:
Would you be open to the idea of accepting a PR that allows you to specify a Rails status code symbol (:ok, :not_found, etc..) but continue outputting the actual status to the docs?

Example:

``` ruby
error 404, 'Missing'
# becomes
error :not_found, 'Missing'
```

A full list can be found [here](http://guides.rubyonrails.org/layouts_and_rendering.html) in section 2.2.13.

iNecas:
Hi, I think it's reasonable.

iNecas:
apipie-rails 0.5.1 released, including fix for this issue

PR #468, opened 2016-06-29
Recognize Rack Symbols as Status Codes

alex-tan:
Addresses #466

### DIFF
--- a/lib/apipie/error_description.rb
+++ b/lib/apipie/error_description.rb
@@ -1,7 +1,6 @@
 module Apipie
 
   class ErrorDescription
-
     attr_reader :code, :description, :metadata
 
     def self.from_dsl_data(args)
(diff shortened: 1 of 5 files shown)


Tybot204:
:+1: Would love to see this merged in. I prefer Rack codes since I think it makes your code much clearer! I usually can't remember all of the codes and their meanings.

lib/apipie/error_description.rb
```diff
@@ -1,9 +1,11 @@
 module Apipie
 
   class ErrorDescription
-
     attr_reader :code, :description, :metadata
 
+    class UnrecognizedCode < Exception
```
iNecas: We have all other errors at one place her https://github.com/Apipie/apipie-rails/blob/296c097f9acb951149d3a67db69b84eb216bdd7e/lib/apipie/errors.rb. Also, prefer `StandardError`, or in this case even `Apipie::Error` over `Exception`

iNecas:
Nice one. One comment inline, but should be good to go otherwise

alex-tan:
@iNecas put the change into place! Thanks for reviewing.

alex-tan:
Just an update: merged master back in and the build is green ✅

iNecas:
Could you rebase you branch, rather than merging the master into your branch: it makes it simpler to test and we prefer rebase over merge in general. This branch has conflicts at the moment.

alex-tan:
@iNecas done!

iNecas:
Thanks @alex-tan

alex-tan:
@iNecas 👍

iNecas:
apipie-rails 0.5.1 released, including this change

Issue #485, opened 2016-08-30
Accessing a non-existent method when config.authorize in't nil


nozik:
Hi,

When I implement config.authorize, but try to access a non existent method (send a random string as params[:method] in apipies_controller), authorize_doc fails because it only checks that Apipie.configuration.authorize isn't nil, but doesn't check if @doc is nil.

Can easily be fixed by checking for nil in authorize_doc.

Cheers

PR #487, opened 2016-08-30
Fix accessing a non-existent method when config.authorize in't nil

nozik:
Related to the issue I've opened - https://github.com/Apipie/apipie-rails/issues/485

### DIFF
--- a/app/controllers/apipie/apipies_controller.rb
+++ b/app/controllers/apipie/apipies_controller.rb
@@ -93,8 +93,8 @@ module Apipie
     end
 
     def authorized_doc
-
-      return @doc if @doc.nil? or not Apipie.configuration.authorize
+      return if @doc.nil?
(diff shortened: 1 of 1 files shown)


app/controllers/apipie/apipies_controller.rb
```diff
@@ -94,7 +94,7 @@ def get_language
 
     def authorized_doc
 
-      return @doc unless Apipie.configuration.authorize
```
iNecas: What about doing this check in separate line like this:

```ruby
return unless @doc
```

It would read a bit easier for me than having both conditions at once.

nozik: done

iNecas:
Thanks @nozik, merged as e0e2b0ea606682c2a48b0bbf34e7ef467942c600

Issue #490, opened 2016-10-13
app_info markup_to_html before translate


belousovAV:
```
translation missing: en.doc.
<p>info</p>
```

Because app_info wrapped `<p>` tag before translate.

PR #491, opened 2016-10-13
fix https://github.com/Apipie/apipie-rails/issues/490

belousovAV:
app_info wrapped `<p>` tag before translate.


iNecas:
Makes sense. Thanks @belousovAV

iNecas:
apipie-rails-0.4.0 has just been released, including this enhancement. Thanks @belousovAV  for help!

Issue #492, opened 2016-10-13
I18n translate at Readme


belousovAV:
Example of translate function for I18n at Readme contained some errors.

if `str = nil ` than `translation missing: en.no key`
if `str = ''`  than `I18n::ArgumentError: I18n::ArgumentError`raised.

PR #493, opened 2016-10-13
fix https://github.com/Apipie/apipie-rails/issues/492

belousovAV:
Example of translate function for I18n at Readme contained some errors.

if `str = nil ` than `translation missing: en.no key`
if `str = ''`  than `I18n::ArgumentError: I18n::ArgumentError`raised.


iNecas:
Makes sense

PR #496, opened 2016-10-18
add IT locale

alepore:
I added the italian locale file


iNecas:
Not speaking italian, but google translate seems legit :) Thanks @alepore

iNecas:
apipie-rails-0.4.0 has just been released, including this enhancement. Thanks @alepore  for help!

PR #501, opened 2016-11-04
Fix type validator message

cindygu-itglue:
The type validation message should have `a` before the type.


mtparet:
Thanks!

iNecas:
apipie-rails-0.4.0 has just been released, including this fix. Thanks @cindygu-itglue  for help!

PR #502, opened 2016-11-05
Boolean Validator uses <code> instead of '

berfarah:
Updated the Boolean Validator to output its message similarly to when you have array validations

This:
```rb
param :value, :bool
```
image

Will now look like this:
```rb
param :value, [true, false, 0, 1]
```
image

when reading the documentation. This makes it clearer that you can pass in the literal values instead of strings of those values.

I can update the ArrayValidator to do the same thing –– I believe it's currently just the Enum one that outputs it with a code.

### DIFF
--- a/lib/apipie/validator.rb
+++ b/lib/apipie/validator.rb
@@ -413,7 +413,8 @@ module Apipie
       end
 
       def description
-        "Must be 'true' or 'false' or '1' or '0'"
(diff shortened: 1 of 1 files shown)


iNecas:
Thanks @berfarah , Merging now

iNecas:
apipie-rails-0.4.0 has just been released, including this enhancement. Thanks @berfarah  for help!

berfarah:
Happy to contribute, even if it was a tiny bit

PR #508, opened 2016-12-15
Adds allow_blank option

MrLeebo:
Based on @Pajk 's comments in the outcome of #403, this is a PR to introduce the `allow_blank: true` parameter option.

Like `allow_nil`, but for blank values. `false`, `""`, `' '`, `nil`, `[]`, and `{}` are all blank.

Usage:

```
param :age, :number, allow_blank: true
```


MrLeebo:
Tests pass on my local machine and the Travis CI error doesn't look like it is related to this. Should I be concerned about that?

mtparet:
Thanks!

iNecas:
apipie-rails-0.4.0 has just been released, including this enhancement. Thanks @MrLeebo  for help!

PR #519, opened 2017-01-27
Replace PNG badges with SVG

olivierlacan:
They're much crisper on high resolution displays. :-)


iNecas:
Yes, the do! Thanks for the tip and caring to fix it @olivierlacan

PR #521, opened 2017-01-30
Allow a blank, not just nil, base_url.

johnnaegle:
I changed my api_base_url from '/api' to '' and had document generation fail because only nil base_urls were allowed.  This relaxes that and allows a blank base_url.


iNecas:
Makes sense. Thanks @johnnaegle

iNecas:
apipie-rails-0.4.0 has just been released, including this fix. Thanks @johnnaegle  for help!

PR #524, opened 2017-02-01
Include delete request parmeters in generated documentation

johnnaegle:
One of my API routes allows multiple objects to be deleted.  In routing, it
looks like this:

```ruby
  resources :posts do
    collection do
        delete :destroy
      end
  end
```

The generated examples from tests did not include the parameters.  This patch
treats :DELETE like PUT/POST/PATCH and includes the parameters in the generated
examples.


iNecas:
Makes sense. Thanks @johnnaegle

iNecas:
apipie-rails-0.4.0 has just been released, including this enhancement. Thanks @johnnaegle for help!

PR #531, opened 2017-03-14
Added recursion for documentation, fixed bug in examples with paperclip

blddmnd:
Right now if you want to use some files from paperclip in your examples (from RSpec) you can receive error:

`'encode': "\x89" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)`

I fixed this bug. Also I added special method for files, so now I can use files in nested hash and this method will convert file data to "FILE CONTENT".


iNecas:
Looks good, thanks @blddmnd

iNecas:
apipie-rails-0.5.0 is out, including this change. thanks @blddmnd

Issue #325, opened 2015-01-27
Respond_to YAML/RAML


baweaver:
While this gem has a lot of potential already in documentation, what would really make it shine is the ability to tie into client generation.

http://raml.org/

If it can export as RAML, you can automatically generate Javascript clients among a few other languages.

Currently I've been musing about how to go about this, so I figured I'd open an issue to gauge interest before I consider trying to work from scratch here.

iNecas:
Hi,

I think we're pretty much in sync with the direction. We currently produce a json for further processing and use is as basic for ruby bindings https://github.com/Apipie/apipie-bindings. They are pretty simple right now, but I would like to add more resource-like experience (I have a POC on that locally, but needs some polishing before getting published). The first step to it was preparation for better testing against real app https://github.com/Apipie/apipie-bindings/pull/22.

Anyway, I know the ad-hoc json format has it's limitation and I would be more than happy to see one that could share an ecosystem of others. Initially, I was thinking about adopting http://apiblueprint.org/. In theory, there should not be problem to support more formats, OTOH the better fit should be done first.

baweaver:
The information's already there for the most part, so I can't see it being too hard a task to manage to port different API handlers. It'd take me a few days to grok the code here and get a better idea of how much work it'd be though.

I'll have to check if RAML has validation support as well as a few other things. I just know it'd make a killer Angular Service factory.

baweaver:
As soon as I have some more time free up, I was planning on trying to do something like this myself, so if you need a hand let me know.

iNecas:
Sounds great, in the meantime, I will try to get my hands dirty with the improving the Ruby bindings. We could then at some point just switch the format in the bindings to use the RAML istead of the custom based…

baweaver:
Right now it's just a matter of understanding how the code works. I might make some type of log as I go through and push it to the Wiki for future reference as well.

iNecas:
Swagger might be another format we could look into http://editor.swagger.io

spinx:
Any progress on this ? I'd like to generate a client in PHP for an API that's documented with apipie.

erlend-sh:
Swagger is now the standardised [OpenAPI specification](https://openapis.org/), so that's a pretty safe bet. Some of the more common complaints about Swagger are directed at the arguably dated swagger-ui project. There's now a modern alternative to this called ReDoc:

http://swagger.io/redoc-openapi-powered-documentation/

[...]

PR #535, opened 2017-03-29
Markdown (APIBlueprint) Support

voltechs:
Hi all,

This is a big PR, but I think it is a good one.

There are definitely some breaking changes in here, so it might need a bigger discussion.

Breaking Changes:
* hash => object
* numeric => number
* errors are now simply a response, the "error-ness" of it is implied via the response code and interpreted by humans.
* DSL changed slightly to be a little more consistent (all responses need a code, second param is the body of the response)

Improvements + Additions
* Broke out large files into sub-files based on classes (see DSL)
* Markdown/APIBlueprint support!!

Happy to discuss these changes :)


iNecas:
Hi,

First of all thanks for doing this: I believe this is an important move that one needed to do for getting apipie to the next level.

Before we go deeper into the discussion, I would suggest investing time to keeping the backward compatibility: apipie has been here for a while and we introduced very little (if any) backward incompatibile changes. If nothing else, we will need a path though deprecations, but I'm more in favor of aliases, unless it harms the code.

This is easier to achieve on DSL level, and could be harder for json output format. In ideal world, we should have a period where we provide both json formats and then we can either: deprecate one over the other; or keep multiple formats and leave it as a benefit to be able to define another formats if needed, and from some disucssion we had previously (https://github.com/Apipie/apipie-rails/issues/325#issuecomment-246730599), there might be other formats in play, such as Swagger, so this option might be the way to go.

micapam:
Looks like this PR has gone stale, but it's a feature I'm keen to see supported. 

Assuming @voltechs is no longer interested in this PR - I could pick it up, potentially? Let me know if there's still interest among the maintainers in this feature.

PR #537, opened 2017-04-05
Do not mutate strings passed as config options, fixes #461

### LINKED ISSUE #461: Can't modify frozen string literal exception
Apiepie appears to destructively modify configuration strings. This is incompatible with the following example file in Ruby >= 2.3:

``` ruby
# frozen_string_literal: true
Apipie.configure do |config|
  config.default_version = '1.0'
end
```

samphilipd:


iNecas:
Thanks @samphilipd

iNecas:
apipie-rails-0.5.0 is out, including this change. thanks @samphilipd!

PR #543, opened 2017-05-05
Updated setting default locale for api documentation

DmitryKK:


iNecas:
What is the issue this PR fixes? Just to understand better the reasoning behind the change?

DmitryKK:
@iNecas, the fact is that configuration `Apipie.configuration.default_locale' is not working. Api documentation shows default language by `I18n.default locale`. You may check

iNecas:
Oh, I see. Makes sense then :) Merging

iNecas:

Haniyya:
This seems to result in a breaking change:
If you set your `default_locale` to something other than 'en' and defined your api, this change will alter all the links in the documentation to use a locale (like `users/1.de.html`) which leads to non-existant documentation errors for everything.
Workaround is to manually remove the locale from the urls, which leads you back to a working state.

PR #552, opened 2017-08-14
Fallback to apipie views when application override isn't present

tstrachota:
Before this change apipie-rails required all views to be copied into the
application's view folder even when only one of the files contained some
modification. After this change it is possible to override just a single
view.


iNecas:
Makes sense, thanks @tstrachota

iNecas:

PR #557, opened 2017-09-07
Fix reloading when extending the apidoc from concern

iNecas:


iNecas:
In #554, we introduce the way to extend the apidoc from concern. There was some issue, including the concerns from to_prepare block. Leading to:

```
Could not find method description for ExtendedController#create. Was the method defined?
...
    config.filter_parameters += [:password]
    config.to_prepare do
      ExtendedController.send(:include, Concerns::ExtendingConcern)
    end
  end
end
```

The code change handles the situation when the controllers get updated during the code reload to make sure we don't loose once-loaded method description that would need to get updated.

ares:
Works absolutely great! Tested code changing in the concern itself as well as in the include target. @inecas += :1st_place_medal:

iNecas:
Apipie-rails-0.5.3 with the fix was just released.

Issue #549, opened 2017-07-27
Language issue


HienNguyenAsnet:
I have language issue when use v0.5.1. I don't config any language for document but when click link on document. I got error
```
Oops!! Method index.en not found for resource events.
```

yakjuly:
same when I use v0.5.2, currently, I configure as below to avoid link issue.

```
config/initializers/apipie.rb
Apipie.configure do |config|
  config.translate         = false
  config.default_locale = nil
  ...
end
```

markmoser:
If I set default_locale to nil in my config file while using `apipie-rails (0.5.4)` the live site works great. However, `rake apipie:static` throws an error:
> Writing examples to a file
> rake aborted!
> I18n::InvalidLocale: nil is not a valid locale

iNecas:
Closing via https://github.com/Apipie/apipie-rails/pull/562

OrrYakobi:
> same when I use v0.5.2, currently, I configure as below to avoid link issue.
> 
> ```
> config/initializers/apipie.rb
> Apipie.configure do |config|
>   config.translate         = false
>   config.default_locale = nil
>   ...
> end
> ```

This should be added to the documentation!

PR #562, opened 2017-10-14
prevent lang in url when config.translate is false

markmoser:
This helps issue #549


iNecas:
Makes sense. Thanks @markmoser

iNecas:
Apipie-rails 0.5.5 with this change just got released. Enjoy and thanks again for contribution

PR #566, opened 2017-10-30
Allow resource level deprecation

cross-p6:


cross-p6:
I intend to submit this as a PR soon™, but I want to test it first. Sorry for the early PR.

PR #567, opened 2017-11-01
Allow for resource-level deprecations

cross-p6:
Resources can now be marked as deprecated. 
This marking is also referenced at the method level, meaning that resources marked as
deprecated will have all their methods marked as such as well.

This commit also includes related updates to the views used for
generating documentation

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -136,6 +136,9 @@ app_info
 meta
   Hash or array with custom metadata.
 
+deprecated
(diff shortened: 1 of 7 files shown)


iNecas:
Thanks. Looks good and makes sense to me.

iNecas:
Apipie-rails 0.5.5 with this change just got released. Enjoy and thanks again for contribution @cross-p6

PR #568, opened 2017-11-05
Added french locale

giglemad:
Some traductions might prove a bit rough to understand or approximate. I am pretty confident it will do its job otherwise.


iNecas:
Looks french :), Thanks @giglemad

Issue #273, opened 2014-08-13
Describing a response


whoisstan:
So i have a regular 'index' method which responds with a list of items. How would I describe the response format? Can't seem to find it.

saneshark:
I'd be open to working on this. I've used JSON-Schema specs in the past which is great because it allows you to also identify attributes that are for example of type String, but could also potentially be null. As a ruby developer we often overlook the value of type handling and I can't think of a better way to define it other than to have some equivalent of request/view specs that compare the response.body to against such a schema. It would be very nice if the example generators could include comments in the JSON perhaps that are based on such a schema spec.

jonricaurte:
@saneshark apipie does support nils. You can use allow_nil: true which can be found here:

https://github.com/Apipie/apipie-rails

jeremywadsack:
:+1: 

An issue with using the `examples` command is that it "will be formatted as code." More specifically, it seems to format as JSON (or something). So when I have `formats ['csv']` it color codes things that shouldn't be colored. Even for JSON response, if my `example` block includes more than just the returned value the formatter tries to make all of that into code. So I have to use `# comments` to add notes to the example.

Would be great to see this enhanced. What about just letting the author use marked-up content for the `examples`, rather than trying to do something fancy (that may not make sense for all cases)?

cgagon:
Are there any plans to undertake this feature?  RAML and Swagger both support this so I think it would be a great addition to the DSL.

santiph:
+1

elasti-ron:
See #588 for a discussion on a possible approach for addressing this topic.

sordina:
This is especially critical given that the Swagger-Codegen community seem have a policy of making generated code return nil for API requests if the response isn't documented...

https://github.com/swagger-api/swagger-codegen/issues/6388#issuecomment-326991057

iNecas:
@sordina please have a look at https://github.com/Apipie/apipie-rails/pull/588 - the implementation is pretty close, any early feedback can help us moving this feature forward.

iNecas:
Fixed by https://github.com/Apipie/apipie-rails/pull/588

PR #569, opened 2017-11-07
Added Swagger generation

elasti-ron:
Hi,

I've added the capability to auto-generate static [Swagger](https://swagger.io/) definition files from the Apipie DSL.

The resulting Swagger ([OpenAPI 2.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md)) compliant JSON file can then be used to auto-generate REST clients, imported into tools like Postman, and so forth.

Swagger generation is implemented by crawling through Apipie's resource_descriptions construct.

The DSL is slightly enhanced to allow indication of default values of optional parameters.

There are a few configuration options that can be used to change the overall structure of the generated swagger file, in order to support both in:formData and in:body REST request styles.

Looking forward to hear your thoughts.

### DIFF
--- a/app/controllers/apipie/apipies_controller.rb
+++ b/app/controllers/apipie/apipies_controller.rb
@@ -14,10 +14,42 @@ module Apipie
       end
     end
 
+
+    def render_swagger_json
+      # the 'authorize' configuration is not supported for swagger output
+      head :forbidden and return if Apipie.configuration.authorize
+
+      @language = get_language
+
+      Apipie.load_documentation if Apipie.configuration.reload_controllers? || (Rails.version.to_i >= 4.0 && !Rails.application.config.eager_load)
+
+      I18n.locale = @language
+
+      prev_warning_value = Apipie.configuration.swagger_suppress_warnings
+      Apipie.configuration.swagger_suppress_warnings = true
+
+      doc = Apipie.to_swagger_json(params[:version], params[:resource], params[:method], @language)
+
+      Apipie.configuration.swagger_suppress_warnings = prev_warning_value
+
+      unless doc
+        render 'apipie_404', :status => 404
+        return
+      end
+
+      render :json => doc
+    end
+
+
     def index
       params[:version] ||= Apipie.configuration.default_version
 
       get_format
+
+      if params[:type].to_s == 'swagger' && params[:format].to_s == 'json'
+        render_swagger_json
+        return
+      end
 
       respond_to do |format|
 
--- a/app/controllers/apipie/apipies_controller.rb
+++ b/app/controllers/apipie/apipies_controller.rb
@@ -15,40 +15,14 @@ module Apipie
     end
 
 
-    def render_swagger_json
-      # the 'authorize' configuration is not supported for swagger output
-      head :forbidden and return if Apipie.configuration.authorize
-
-      @language = get_language
-
-      Apipie.load_documentation if Apipie.configuration.reload_controllers? || (Rails.version.to_i >= 4.0 && !Rails.application.config.eager_load)
-
-      I18n.locale = @language
-
-      prev_warning_value = Apipie.configuration.swagger_suppress_warnings
-      Apipie.configuration.swagger_suppress_warnings = true
-
-      doc = Apipie.to_swagger_json(params[:version], params[:resource], params[:method], @language)
-
-      Apipie.configuration.swagger_suppress_warnings = prev_warning_value
-
-      unless doc
-        render 'apipie_404', :status => 404
-        return
-      end
-
-      render :json => doc
-    end
-
-
     def index
       params[:version] ||= Apipie.configuration.default_version
 
       get_format
 
       if params[:type].to_s == 'swagger' && params[:format].to_s == 'json'
-        render_swagger_json
-        return
+        head :forbidden and return if Apipie.configuration.authorize
+        should_render_swagger = true
       end
 
       respond_to do |format|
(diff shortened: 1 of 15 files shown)


lib/apipie/swagger_generator.rb
```diff
@@ -0,0 +1,583 @@
+module Apipie
+  module Validator
+    class EnumValidator < BaseValidator
```
iNecas: Why not putting it directly to the validator definition?

lib/apipie/swagger_generator.rb
```diff
@@ -0,0 +1,583 @@
...
+    # result from changes to the input and those that result from changes to the generation
+    # algorithms.
+    # note that at the moment, this only takes operation ids into account, and ignores parameter
+    # definitions, so it's only partially useful.
+    def include_op_id_in_master_identifier(op_id)
+      @master_id = Zlib::crc32("#{@master_id} #{op_id}")
+    end
+
+
+    def save_and_compare
```
iNecas: This method is not being called from any place? Should it be exposed from the rake?

lib/apipie/swagger_generator.rb
```diff
@@ -0,0 +1,583 @@
...
+
+    # the @master_id is a number that is uniquely derived from the list of operations
+    # added to the swagger definition (in an order-dependent way).
+    # it is used for regression testing, allowing some differentiation between changes that
+    # result from changes to the input and those that result from changes to the generation
+    # algorithms.
+    # note that at the moment, this only takes operation ids into account, and ignores parameter
+    # definitions, so it's only partially useful.
+    def include_op_id_in_master_identifier(op_id)
+      @master_id = Zlib::crc32("#{@master_id} #{op_id}")
```
iNecas: we should probably require `zlib` somewhere: it's probably loaded by rails somewhere, but just to be on safe side

lib/apipie/swagger_generator.rb
```diff
@@ -0,0 +1,583 @@
...
+
+    def info(msg)
+      print "--- INFO: [#{ruby_name_for_method(@current_method)}] -- #{msg}\n"
+    end
+
+    def investigate(msg)
+      # print "??? RESEARCH: [#{ruby_name_for_method(@current_method)}] -- #{msg}\n"
+    end
+
+    def dump_param_desc(param_desc)
```
iNecas: Unused method?

iNecas:
There are some unused methods and commented code, that should be cleaned first. I have not tested this yet, but my view on this is we could get this in pretty soon, as it has not influence on other features, and can be pretty useful for many folks.

iNecas:
I wonder if we could have a test on checking that the swagger output valid, in the tests.

iNecas:
Another question: would it make sense to expose this via an `/apidoc` endpoint, similarly as we have with `apidoc.json`?

elasti-ron:
Thanks @iNecas  - good points.  I'll push a cleaned-up version early next week.

Re exposing via `/apidoc`:  yes, would be useful.  Would probably make sense to auto-fill the swagger `host` field with the hostname/port from the request object.  I'll look into this.

Re testing the validity of the swagger output: would be great to add. A possible approach might be to validate against the [Swagger schema](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v2.0/schema.json) (perhaps by using the [json-shema](https://github.com/ruby-json-schema/json-schema) gem?) on the output. What do you think?

iNecas:
Agreed with all you said :)

elasti-ron:
Pushed a cleaned-up version addressing the code issues you highlighted.

Will keep you posted on progress re the `/apidoc` endpoint and the swagger schema validation.

elasti-ron:
Hi @iNecas - it turns out that rendering at runtime from `/apidoc` is not as simple as I had hoped.  It will take me some time before I'm able to look into this further, and I'll only be able to continue working on this (and on the swagger schema validation) next week.

If you feel comfortable with what I've posted so far as a first version, it might be best to merge that into the master branch and not wait for the additional functionality.  This will reduce the risk of incompatibilities creeping into the code.

iNecas:
I'm ok with waiting for the `/apidoc` endpoint, but I wukd lje to have the schema validator test before merging this to keep it working with the further updates

elasti-ron:
The schema validation was surprisingly easy to add.  Pushed a version that includes it.

elasti-ron:
ping

elasti-ron:
Hi @iNecas - runtime swagger generation is now supported (with some limitations - see the README).

app/controllers/apipie/apipies_controller.rb
```diff
@@ -14,11 +14,43 @@ def authenticate
...
     end
 
+
+    def render_swagger_json
+      # the 'authorize' configuration is not supported for swagger output
+      head :forbidden and return if Apipie.configuration.authorize
+
+      @language = get_language
+
+      Apipie.load_documentation if Apipie.configuration.reload_controllers? || (Rails.version.to_i >= 4.0 && !Rails.application.config.eager_load)
```
iNecas: There is quite a lot of code duplicated from the `index` method. Could you look into how to DRY this a bit?

Otherwise, the implementation looks good and I think we are close to merging this

lib/apipie/configuration.rb
```diff
@@ -165,6 +170,12 @@ def initialize
       @translate = lambda { |str, locale| str }
       @persist_show_in_doc = false
       @routes_formatter = RoutesFormatter.new
+      @swagger_content_type_input = :form_data  # this can be :json or :form_data
+      @swagger_json_input_uses_refs = false
+      @swagger_include_warning_tags = false
+      @swagger_suppress_warnings = false #[105,100,102]
```
iNecas: What does ` #[105,100,102]` mean?

iNecas: Nm: I looked at the readme and here we go :)

iNecas:
Thanks @elasti-ron, I don't see any other reason why holding this in a PR any longer. Thanks for hard work and this one. I'm sure it will be useful for others as well

elasti-ron:
Thanks @iNecas - it was good to collaborate with you on this.

As a next step, I'd like to add support for describing response structures, as it would make the output much more useful.

I've written a [proposal](https://gist.github.com/elasti-ron/e6cb880fdbcf855ef22143021a435bbf) to describe my suggested approach, and I would highly appreciate your feedback on it.

iNecas:
Thanks for moving apipie forward. I would have few comments to the proposal: would you mind opening the proposal as a PR, with the markdown document in it. The would for the PR would not be to merge that, but to have a meaningful discussion over the diff itself.

In general, I like the idea of combining the concepts we already have in the apipie. Although, I would suggest introducing additional keyword (something like `property`) to the dsl: when defining response only keys, I'm not sure it makes sense to talk about `parameter` in that regard. It would be more like an alias.

I will have more comments, but I would like to leave that for a new PR to continue with the discussion.

iNecas:
Also, make sure to advertise this in https://github.com/Apipie/apipie-rails/issues/273, as there are other folks interested into this functionality, that could have a valuable input to the progress of this.

elasti-ron:
Done.  #588.

PR #571, opened 2017-11-12
Prevent missing translation span in title

mbacovsky:
When the translation is missing it is wrapped with
<span class="translation_missing"... by rails. HTML elements
are not allowed within the title tag.

### DIFF
--- a/app/views/layouts/apipie/apipie.html.erb
+++ b/app/views/layouts/apipie/apipie.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title><%= t('apipie.api_documentation') %></title>
+  <title><%= t('apipie.api_documentation', :default => 'Api Documentation') %></title>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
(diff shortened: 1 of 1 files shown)


iNecas:
I'm not sure this is the right fix, as for other translations would also get the `missing_translation` error. In the case of `api_documentation`, the default generated string is `Api Documentation`, which looks quite good. however, for other keys, such as `apipie.no_docs_found_descr`, it would give you `No Docs Found Descr`, instead of the english version.

I would recommend setting up the fallbacks in I18n for the rails app itself, such as here https://github.com/theforeman/foreman/pull/5034

mbacovsky:
@iNecas, I like the global fallbacks you are suggesting. It may not be the best solution for other projects using Apipie so I'd leave it on those projects and keep up with the rails default beahvior. 

By default the I18n is producing HTML marked strings instead of plain text for missing translations (in erb) which is not allowed in `<title>`.  The patch is handling just this case where the default makes no sense. I think it may be useful to get in both solutions. Makes sense?

iNecas:
Not opposed to that.

iNecas:
Thanks @mbacovsky

Issue #551, opened 2017-08-09
api! break apipie


phspies:
When using api :method, "<route>" in a controller all works fine. The moment I try to use api!, the following error is seen "uninitialized constant OculusController"

Haniyya:
@phspies Had the same problem. My problem was that i had 'dangeling routes' in `conf/routes.rb`. 
Something like `resources :oculus` or something might create routes to controllers that don't exist (anymore).

So the solution is to either delete those routes in `routes.rb` or not to use `api!`.

Haniyya:
That said, it would probably be a good idea to catch those errors somewhere since this is not really an informative or expected error.

Haniyya:
I guess this could be closed now.

iNecas:
yes, thanks for heads up

PR #575, opened 2017-11-30
Replaced String#constantize with String#safe_constantize so apipie won't break on a missing constant

Haniyya:
As mentioned in #551 apipie has a tendency to break if there are routes pointing to missing controllers.
This PR makes apipie skip these routes, thereby ignoring the missing controllers.


iNecas:
Thanks @Haniyya

Issue #577, opened 2017-12-07
Ruby style guide compliance


Haniyya:
Not much of an issue just an Idea to let [rubocop](https://github.com/bbatsov/rubocop) autocorrect the codebase (or better yet, make a task in the travis-ci) to help improve the codebase.

## Pros
1. Better readability
1. Slight performance improvements in ruby >= 2.3 (frozen_string_literals and such)
1. It would be easier for contributors to navigate the codebase if it followed some common standards.
1. Pretty easy to implement (on a surfice level at least) 

## Cons
1. Big, unwieldy PR
1. No short term benefit
1. Previous contributors/maintainers may be irritated

Opinions?

PR #578, opened 2017-12-07
Ruby style guide compliance

Haniyya:
Example implementation of #577 .

See issue for more info.


mathieujobin (changes requested):
there is too many changes in this PR 

lets address styling issue one at the time so its easier to review
and safe to merge

PR #595, opened 2018-01-16
WIP: Feature/better example display

Haniyya:
This adds and changes a few things:

1.   You can now add a `doc_title` metaparameter to your RSpec-examples:
```ruby
  it 'just works!', doc_title: 'Thing that works.' do
    expect(it).to work
  end 
```
  Which will then be used as the title of an example in the documentation.
2. On :POST and :PUT requests, the params will be displayed as an easily readable JSON object in the documentation.
3. Examples are collapsible and scrollable (response data is anyway).


Haniyya:
WIP since i probably need to fix/write some tests still.

iNecas:
Would it be possible to split the PR into some smaller isolated PRs that could be reviewed separately? Big PRs are usually harder to tackle.

Haniyya:
@iNecas I could split it into 3 that correlate to the three features mentioned above if thats okay.

iNecas:
Yes please

Haniyya:
@iNecas I tried to isolate the changes as best i could but #598 still turned out to be quite big. This is mostly down to the changes in markup that are necessary to have any effect on the user experience.

Issue #445, opened 2016-03-25
Add "title" to json example recordings


danielkummer:
Hi there,
according to the readme, it should be possible to add the title attribute to examples:

> You can add a title to the examples (useful when showing more than one example per method) by adding a 'title' attribute.

I tried to find out, how this can be done using rspec - I don't want to do it manually when using multiple rspec examples. Can it be done using rspec metadata?
e.g
`
    it 'does something', :show_in_doc, title: 'This is the example title' do
        ...
    end
`

Kind regards

jpletner:
@danielkummer Did you ever figure out how to add a title? I tried using rspec metadata unsuccessfully.

johnnaegle:
I didn't see a way to do this, but if you add the title attribute in generated JSON example, it is persisted.  The old example file is read in and merged with the new file.  YMMV.  I would also like to have the title come in as rspec metadata.

IlyaBylich:
I'm using something like:
``` ruby
# spec/support/apipie_helper.rb (must be required in your spec_helper.rb)
module ApipieRecorderPatch
  def record
    super.try(:merge, title: RSpec.current_example.metadata[:doc_title] || 'Default')
  end
end

class Apipie::Extractor::Recorder
  prepend ApipieRecorderPatch
end
```

Later you can use it by writing
``` ruby
it 'returns a valid response', doc_title: 'Sample #1' do
  # your test
end
```

johnnaegle:
@iliabylich Thanks! Worked great for me.

PR #597, opened 2018-01-17
RSpec doc title

Haniyya:
This enables the extractor to record `doc_title`s from an RSpec example.
This information is currently not used but it will be in #595 .

The changes are heavily inspired by #445 and @ghost s solution.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1401,6 +1401,19 @@ And then configure RSpec in this way:
 This way, when running in recording mode, only the tests that have been marked with the
 ``:show_in_doc`` metadata will be run, and hence only those will be used as examples.
 
+If you want to distinguish your recorded examples from one another, you can provide a ``doc_title``
+that will be displayed in the documentation later, like this:
+
+.. code:: ruby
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -1401,6 +1401,19 @@ And then configure RSpec in this way:
 This way, when running in recording mode, only the tests that have been marked with the
 ``:show_in_doc`` metadata will be run, and hence only those will be used as examples.
 
+If you want to distinguish your recorded examples from one another, you can provide a ``doc_title``
```
iNecas: It should be mentioned that this works only with RSpec. The ultimate patch would be to make it work also with minitest, in case https://github.com/ordinaryzelig/minispec-metadata/blob/master/lib/minispec-metadata/it.rb#L30 is present.

mathieujobin (changes requested):
as @iNecas said... 

I will close in a few months if no updates

mathieujobin (approved):
finally hahaha

PR #598, opened 2018-01-17
Request data as json

Haniyya:
This PR changes the appearance of examples and allows for `request_data` of `:PUT :PATH :POST` requests to be displayed as json. Which for a lot of people is easier than the standard query-string.
This is part and precursor of #595 which fixes some of the markup.

A few changes are made here:
1. Examples are no longer pre-formated when passed to the template. Instead the normal case will be to have Hash-examples that have their own partial and a seperate one for simple String examples.
2. The `@params` attribute of the `Extractor::Recorder` contains more information about the request and is copied to `request_data_json` when serializing. This field stays untouched until it reaches the template (i didn't use `params` here since it was being modified on deserialization).
3. The included Bootstrap was updated to 3.3.7.
4. There is (incomplete and not working, fixed in #595) markup for method display.


lib/apipie/extractor/recorder.rb
```diff
@@ -109,7 +111,8 @@ def record
            :path => @path,
            :params => @params,
            :query => @query,
-           :request_data => @request_data,
+           :request_data => request_data,
```
iNecas: Why removing the `@` here? Doesn't seem like this could work?

lib/apipie/extractor/recorder.rb
```diff
@@ -109,7 +111,8 @@ def record
...
+           :request_data_json => @params,
```
iNecas: How does this differ from `:params` key?

lib/apipie/method_description.rb
```diff
@@ -194,7 +191,12 @@ def load_recorded_examples
         find_all { |ex| ex["show_in_doc"].to_i > 0 }.
         find_all { |ex| ex["versions"].nil? || ex["versions"].include?(self.version) }.
         sort_by { |ex| ex["show_in_doc"] }.
-        map { |ex| format_example(ex.symbolize_keys) }
+        map do |ex|
+        %w(request_data response_data request_data_json).each do |key|
```
iNecas: Beware of indentation

Haniyya:
@iNecas Thanks for taking a look! I'll adjust this once I'm off work :)

robikovacs:
Are any news on this PR?

sasharevzin:
Any updates on it?

sharkey11:
Any updates???

Haniyya:
I'm sorry but I don't have access to the fork here since I don't work at the company I originally did this for any longer. I guess you could take the changes made here and open a new fork, based on the current master, but I lack the time (and frankly, the understanding of apipie since I've not been working with it for over 2 years) to continue.

sasharevzin:
@Haniyya I understand. did you find the replacement for apipie?

Haniyya:
@sasharevzin My previous company worked with apipie until the end (more or less). Right now I mainly work on graphql or thrift apis, which makes apipie kind of unnecessary? (I mean there is still value in the way you can document everything and provide examples easily, but the need rarely arises beyond a typed interface). 

I really enjoyed working with apipie while I did and it was a great way to document a legacy rest api peace by peace so to speak, but in greenfield projects we usually alredy start with some standard like thrift, graphql or OpenApi.
And thank you for understanding. I really do feel bad leaving this feature as it seems close to being done, but I can't get my head around it anymore.

mathieujobin:
reopen if you have time to resolve the conflicts and adjust to today's situation

PR #600, opened 2018-01-18
Gzip examples

Haniyya:
This PR adds the option (default: false) to compress recorded examples using `Zlib`.
This can be benefitial by keeping examples out of git diffs and in reducing their size for big test suits.


iNecas:
Makes sense Thanks @Haniyya  for nice PR, bonus points for tests!

iNecas:

PR #605, opened 2018-02-14
Use SHA1 instead of MD5 to enable using APIPIE at FIPS-enables systems

iNecas:
FIPS forbits using MD5 at system level, therefore we need to use
something else inside.


iNecas:
Seems no one objects this change. Merging now.

PR #608, opened 2018-03-10
fix bug allow_blank=>false doesn't work

genkin-he:
fix bug :allow_blank=>false doesn't work


iNecas:
Seems the travis is failing on Rails 5+

mathieujobin:
This problem is solved in 0.7.x

PR #613, opened 2018-04-12
Use underscore when appending `controller` to the default controller string

kevinmarx:
The current implementation causes namespaced controllers to fail when using the `api!` dsl method. This is due to `route.defaults[:controller]` being `'api/resource'` which then when `Controller` is added and `camelize` called will result in `Api::Resourcecontroller` instead of `Api::ResourceController`. By appending controller using snake case first results in `_controller` being properly camelized.


iNecas:
Makes sense. Thanks @kevinmarx for the patch!

iNecas:

PR #614, opened 2018-04-17
Expect :number to have type 'numeric'

akofink:
WIP because I'm still testing

### DIFF
--- a/lib/apipie/validator.rb
+++ b/lib/apipie/validator.rb
@@ -414,6 +414,10 @@ module Apipie
         "Must be a number."
       end
 
+      def expected_type
+        'numeric'
+      end
+
       def self.validate(value)
         value.to_s =~ /\A(0|[1-9]\d*)\Z$/
       end
--- a/spec/lib/validator_spec.rb
+++ b/spec/lib/validator_spec.rb
@@ -42,6 +42,12 @@ describe Apipie::Validator do
 
     end
 
+    describe 'NumberValidator' do
+      it 'should expect a Numeric type' do
+        validator = Apipie::Validator::BaseValidator.find(params_desc, :number, nil, nil)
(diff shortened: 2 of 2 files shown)


akofink:
@mbacovsky this isn't working for me. Care to try it out?

```rb
Apipie.configuration.use_cache = false
Apipie.reload_documentation
Apipie.to_json("v2")[:docs][:resources]['activation_keys'][:methods][0][:params][0]

=> {:name=>"organization_id",
 :full_name=>"organization_id",
 :description=>"\n<p>organization identifier</p>\n",
 :required=>true,
 :allow_nil=>false,
 :allow_blank=>false,
 :validator=>"number.",
 :expected_type=>"string",
 :metadata=>nil,
 :show=>true,
 :validations=>[]}
```

Here's the [param definition](https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/activation_keys_controller.rb#L20).

mbacovsky:
@akofink, I finaly got to this. Pry revealed that there is another NumberValidator in Katello project [1] that takes precedence during validator look up. From the git history it seems the NumberValidator was added to Katello first and then added also to the Apipie.

:+1: from me for this patch. Thanks for the test.

To fix the original issue the redundant validator in Katello needs to be fixed or better removed from Katello completely to use the Apipie's validator with this patch.

[1] https://github.com/Katello/katello/blob/master/lib/katello/apipie/validators.rb#L32

akofink:
Yay! Thank you for taking a look. I'm not sure why I didn't realize the validator was coming from Katello. Doh! ;)

akofink:
@iNecas Could we please have a release with this change included?

iNecas:
Apipie-rails-0.5.9 with this change has been pushed to Rubygems.

PR #615, opened 2018-04-26
Swagger json includes apipie's description as swagger's description

gogotanaka:
Hi! Thank you for maintaining the greate gem.

## Background
When generating a static Swagger JSON, we are missing `description` parameter.

## What this PR for
To pass apipie's description annotation to swagger's description key.


iNecas:
Makes sense, Thanks @gogotanaka

iNecas:

PR #619, opened 2018-05-08
Support response validation

abenari:
This pull request contains 3 patches:
1. Add support for response validation.
With this patch it is easy to validate the response in unit tests with the documentation schema.
It help creating and maintaining a valid response documentation.

2. Supporting specification of array properties with non-string types

3. Allows creating a schema where objects are defined once and referenced by multiple methods.
When generating a client from the schema, the client code becomes easier to use.

PR #621, opened 2018-05-08
Objects as reference in schema

abenari:
This patch allows creating a schema where objects are defined once and referenced by multiple methods.
When generating a client from the schema, the client code becomes easier to use.


abenari:
depends on code from #619. Closing after I have moved the commits into the other pull request.

PR #625, opened 2018-06-25
Allow HTTP and HTTPS schemas in Swagger json

jlebrijo:
This PR includes both schemes in Swagger json.

### DIFF
--- a/lib/apipie/swagger_generator.rb
+++ b/lib/apipie/swagger_generator.rb
@@ -70,6 +70,7 @@ module Apipie
               "x-copyright" => Apipie.configuration.copyright,
           },
           basePath: Apipie.api_base_url(version),
+          schemes: [ "http", "https" ],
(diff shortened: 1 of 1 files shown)


lib/apipie/swagger_generator.rb
```diff
@@ -70,6 +70,7 @@ def init_swagger_vars(version, lang, clear_warnings=false)
               "x-copyright" => Apipie.configuration.copyright,
           },
           basePath: Apipie.api_base_url(version),
+          schemes: [ "http", "https" ],
```
iNecas: I wonder if this shouldn't be configurable? I guess not everyone is exposing their APIs via HTTP (if I understand correctly the reason for this change)

danleyden:
#710 make this configurable and secure by default

mathieujobin:
#710  will get merged instead

PR #626, opened 2018-07-24
Allow ArrayValidator to use other validators to validate array elements

micapam:
I had a situation where I needed to pass an array of integers in the query string. 

I tried using `array_of: Integer` but this didn't work, because the values was passed in as an array of strings: `["1", "2"]` instead of `[1, 2]`. 

This PR fixes that by allowing ArrayValidator to use other validators to validate the elements of the parameter array. So I was able to use `array_of: :number`. It should work for other validators too because it looks them up with `BaseValidator.find`. 

I also added a spec.

### DIFF
--- a/lib/apipie/validator.rb
+++ b/lib/apipie/validator.rb
@@ -225,7 +225,13 @@ module Apipie
 
       def has_valid_type?(value)
         if @items_type
-          value.kind_of?(@items_type)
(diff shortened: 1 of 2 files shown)


iNecas:
Nice and simple. Thanks @micapam

iNecas:
apipie-rails-0.5.10 is all yours :)

micapam:
Cheers @iNecas !

Issue #632, opened 2018-10-09
Allow resources or methods to be tagged with multiple swagger tags


enrique-guillen:
The [`swagger-editor`](https://swagger.io/docs/open-source-tools/swagger-editor) allows clicking on tags and getting all endpoints tagged with that tag. [It seems like](https://github.com/Apipie/apipie-rails/blob/61ad1e69fc98ee248f3f1181830a40d53aae7d41/lib/apipie/swagger_generator.rb#L227) Apipie only supports one customizable tag, which I'm guessing is the name of the class or the `name` in the `resource_description`.

The DSL could work like this:

```ruby
  resource_description do
    name 'Foo | Bar'
    tags 'Foo', 'Bar', 'Baz'
    short 'A bar in the foo namespace that is related to Bazes'
  end
```

And/or:

```ruby
  api :POST, 'foo/bars'
  param :relationships, desc: 'A qux object related to the bar can be created here.'
  resource_tags 'Foo', 'Bar', 'Baz', 'Qux'
  def index; ...; end
```

enrique-guillen:
PR no longer a WIP.

iNecas:
Closed by #634, thanks @enrique-guillen

PR #634, opened 2018-10-13
Adds swagger tags in swagger output

enrique-guillen:
Addresses issue #632, by adding a single `#tags` DSL method.

Has a `swagger_dsl_spec` example now (among others), and the tags are queried for in `SwaggerGenerator` (see `response_validation_spec`).


enrique-guillen:
I wonder what's up with the "random" successes. CI failed a pair of times like the other 3 recent PRs, on the same example for the same Ruby & Rails combination.

iNecas:
Thank you @enrique-guillen, I've seen this errors as well: will need to investigate furter, but I'm pretty sure it's not related to this PR. I'm happy with the quality and thanks for the tests - it makes makes it much easier to accept the changes.

iNecas:
apipie-rails-0.5.11 was just pushed to rubygems, including this change

Issue #623, opened 2018-06-15
Response description isn't beign shown in docs


rodrigovrm:
Hello,

I'm trying to use the response description (`returns` method) but it isn't showing in the method details page.

Actually, after inspecting the view I noticed that it may be intended as it doesn't renders the responses:

https://github.com/Apipie/apipie-rails/blob/master/app/views/apipie/apipies/_method_detail.erb

I was going to fork it and customise this view, but there isn't values for `:returns` in the `method` Hash.

So, is it a bug? Where are responses supposed to be shown? 

Thank you !

cegue:
I am not sure if this is the same issue I am facing but sounds like it is the same so I will give it a try here before creating a new ticket by adding some details.

According with documentation [#response-description](https://github.com/Apipie/apipie-rails#response-description) specifically with:
>The response from an API call can be documented by adding a returns statement to the method description.

I would expect the response, e.g. JSON schema response to be shown in the correspondent method documentation but that is not the case after following examples at [#response-description](https://github.com/Apipie/apipie-rails#response-description). Also, tried increasing verbosity but I was unable to catch any errors upon request.

Any ideas ?

```
Rails 5.2.0
apipie-rails (0.5.9)
```

FlowerWrong:
```
Rails 5.2.0
apipie-rails (0.5.9)
```

@cegue 👍

micapam:
I see some discussion of this issue here, looks like it's a bug: https://github.com/Apipie/apipie-rails/pull/588

micapam:
Looking at the code, looks like it's not so much a bug as a missing feature. 

It's a bit trickier to render the return structure than the params because the params are always a hash, so they can be rendered as a flat two-column table with the nesting implied by the param names (`foo[bar]`). The return is harder, because it can be either a hash or an array. So - although params and returns are pretty similar - we can't just reuse the `_params` partial.

I haven't quite figured out what it should look like, or I'd have a go at writing it. I tried looking at other api doc systems to see how they solve the problem, haven't got it worked out yet.

maykazw:
any info ?

PR #588, opened 2018-01-02
For discussion: proposal for describing responses

elasti-ron:
This pull request is for supporting a discussion, not for merging into the codebase (as per [request](https://github.com/Apipie/apipie-rails/pull/569#issuecomment-354745916) from @iNecas).

Also relevant to #273

PR #635, opened 2018-10-15
fix returns displaying

X1ting:
Issues: #623 #588


iNecas:
Very good. Thanks @X1ting!

iNecas:
Released in apipie-rails 0.5.12

PR #636, opened 2018-10-16
Update sample code in README

pinewong:


iNecas:
Nice catch, thanks @pinewong

PR #637, opened 2018-10-17
[Bugfix] Stop swagger generation from crashing (if a controller's parent doesn't define a resource_description block)

enrique-guillen:
The original feature release caused Apipie's swagger generator (`rake apipie:static_swagger_json`) to crash with this error:

```ruby
NoMethodError: undefined method `_tag_list_arg’ for nil:NilClass
```

Unless you happened to define a `resource_description` for the parent class of a controller. (It turns out the `parent` variable isn't always a resource-desc object.)

This makes sure `_tag_list_arg` is called on non-nil objects only. I marked this as a bugfix because the error message you get doesn't explain how to fix it, and we shouldn't have to define `resource_description` for the parent classes.

`0.5.11` should be marked as faulty. Sorry about that.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -283,6 +283,8 @@ Example:
         property :enum1, ['v1', 'v2'], :desc => "One of 2 possible string values"
       end
    end
+   tags %w[profiles logins]
+   tags 'more', 'related', 'resources'
    description "method description"
    formats ['json', 'jsonp', 'xml']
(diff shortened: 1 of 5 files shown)


iNecas:
apipie-rails-0.5.13 with this fix pushed to rubygems. Thanks @enrique-guillen

PR #640, opened 2018-10-20
Fixes #629

### LINKED ISSUE #629: Response example generation
On APIPie 0.5.10, I find an issue with the automated generation of response examples.

If I modify my `rails helper` to have
```
config.filter_run(show_in_doc: true) if ENV['APIPIE_RECORD']
```
and annotate selected controller specs with `:show_in_doc`
```
it 'has status 404', :show_in_doc do
  get :index, params: params
  expect(response).to have_http_status(:not_found)
end
```
then, as expected, running
```
APIPIE_RECORD=examples bundle exec rake spec
```
executes only those specs so annotated.  But, in spite of a reassuring `Writing examples to a file` the file `doc/apipie_examples.json` is not modified.  If i delete the file then running the specs does not create it. 

If I comment out the `filter_run` then the file is created/modified, but of course it includes "too many" examples, so I have to manually edit the file to pick out the examples to use, which makes me sad.

Any ideas?

Marri:
The `FunctionalTestRecording` module's version of `process()` was not being called when running tests, which meant `@collector` was never initialized and so examples were not written to the file.
To fix this, we need to prepend `FunctionalTestRecording` to `ActionController::TestCase::Behavior`, not `ActionController::TestCase`, for reasons which I honestly did not manage to determine. (You can follow some of my debug rambling in the original issue.)

I'm not actually clear if this fix is safe to merge as is; presumably the original version is there for a reason, I'd've expected the `prepend()` to apply to the subclass as well as the parent, and this may be more version-specific than I'm giving it credit for. My testing was all done against APIPie 0.5.13 and Rails 5.1.6 and Ruby 2.5.1 and RSpec 3.7.2. The full codebase is available at [https://github.com/Marri/glowfic](https://github.com/Marri/glowfic).

### DIFF
--- a/lib/apipie/extractor.rb
+++ b/lib/apipie/extractor.rb
@@ -23,6 +23,7 @@ class Apipie::Railtie
       end
     else
       ActionController::TestCase.send(:prepend, Apipie::Extractor::Recorder::FunctionalTestRecording)
+      ActionController::TestCase::Behavior.send(:prepend, Apipie::Extractor::Recorder::FunctionalTestRecording)
     end
   end
 end


lib/apipie/extractor.rb
```diff
@@ -22,7 +22,7 @@ class Apipie::Railtie
         prepend Apipie::Extractor::Recorder::FunctionalTestRecording
       end
     else
-      ActionController::TestCase.send(:prepend, Apipie::Extractor::Recorder::FunctionalTestRecording)
+      ActionController::TestCase::Behavior.send(:prepend, Apipie::Extractor::Recorder::FunctionalTestRecording)
```
iNecas: The new version is not working for me. So I guess the original line was there for a reason, Why not to keep both lines in, so that it works for everyone?

iNecas:
Test failure unrelated. Thanks @Marri

iNecas:

Marri:
Thank you!

PR #642, opened 2018-11-09
Support for adding new params via `update_api`

iNecas:
Before this patch, we only supported updating nested hashes. For example

      param :user, Hash do
        param :login, String
      end

and

      param :user, Hash do
        param :oauth, String
      end

has been merged properly. However when merging top-level params, such as

      param :login, String

and

      param :oauth, String

no changes were propagated.


iNecas:
As a bonus, I've also fixed the intermittent test failure we've seen in master for past several weeks. @iNecas is ready for the weekend :)

iNecas:

PR #645, opened 2018-11-15
HTML escape validator values

ekohl:
This fixes an issue when a validator includes a & that needs to be escaped in HTML.

### DIFF
--- a/spec/lib/validator_spec.rb
+++ b/spec/lib/validator_spec.rb
@@ -50,12 +50,64 @@ describe Apipie::Validator do
     end
   end
 
+  describe 'BooleanValidator' do
+    it "should validate by object class" do
+      validator = Apipie::Validator::BooleanValidator.new(params_desc)
+      expect(validator.validate("1")).to be_truthy
+      expect(validator.validate(1)).to be_truthy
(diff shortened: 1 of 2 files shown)


spec/lib/validator_spec.rb
```diff
@@ -50,12 +50,64 @@
...
+    it "should have a valid description" do
+      validator = Apipie::Validator::BooleanValidator.new(params_desc)
+      expect(validator.description).to eq('Must be one of: <code>true</code>, <code>false</code>, <code>1</code>, <code>0</code>.')
+    end
+  end
+
   describe 'ArrayClassValidator' do
     it "should validate by object class" do
-      validator = Apipie::Validator::ArrayClassValidator.new(params_desc, [Fixnum, String])
+      validator = Apipie::Validator::ArrayClassValidator.new(params_desc, [Integer, String])
```
iNecas: This actually fails the tests: let's keep this change out of this PR

iNecas:
All nice and :green_heart:. Thanks for adding additional examples in tests. Merging now and a new release will land today

iNecas:

PR #646, opened 2018-11-15
Do not fail if JSON examples file could not be parsed

dLobatog:


iNecas:
What was the reason the examples could not be parsed? I'm a bit concerned that people will actually not notice the log message, and they will just miss that there is something wrong with the examples in the first place, and might discover that lately just by accident.

PR #648, opened 2018-12-02
Use configured value when swagger_content_type_input argument is not passed

nullnull:
Hi.
I think it would be better to use configured `swagger_content_type_input` value as default value of `rake apipie:static_swagger_json` option.
Or at least I'd like to update README.md. What do you think?


iNecas:
Makes sense. Thanks @nullnull

iNecas:
apipie-rails-0.5.15 including this change has just been released

Issue #654, opened 2018-12-13
When using Apipie.configuration.authorize: new_doc[:docs][:resources] is an array instead of a hash when accessing docs causing errors:


NielsKSchjoedt:
When I use `config.authorize` then any documentation page (other than the `config.doc_base_url`) fails.

When accessing the documentation index page, it works: `localhost:3000/service_api_doc`, but when e.g. accessing one of the links (e.g. localhost:3000/service_api_doc/1/users.en.html) to a documentation for a controller, I get the following: 

```
I, [2018-12-13T20:11:55.366951 #48414]  INFO -- : [89a49504-ca95-4cfb-b5d7-024ec4f3e677] Processing by Apipie::ApipiesController#index as HTML
I, [2018-12-13T20:11:55.367054 #48414]  INFO -- : [89a49504-ca95-4cfb-b5d7-024ec4f3e677]   Parameters: {"version"=>"1", "resource"=>"users.html"}
I, [2018-12-13T20:12:00.758601 #48414]  INFO -- : [89a49504-ca95-4cfb-b5d7-024ec4f3e677] Completed 500 Internal Server Error in 5391ms (ActiveRecord: 11.6ms)
F, [2018-12-13T20:12:00.760367 #48414] FATAL -- : [89a49504-ca95-4cfb-b5d7-024ec4f3e677] 
NoMethodError (undefined method `[]' for nil:NilClass):
  apipie-rails (0.5.14) app/controllers/apipie/apipies_controller.rb:120:in `block in authorized_doc'
  apipie-rails (0.5.14) app/controllers/apipie/apipies_controller.rb:118:in `select'
  apipie-rails (0.5.14) app/controllers/apipie/apipies_controller.rb:118:in `authorized_doc'
  apipie-rails (0.5.14) app/controllers/apipie/apipies_controller.rb:51:in `block in index'
  actionpack (4.2.11) lib/action_controller/metal/mime_responds.rb:212:in `respond_to'
  apipie-rails (0.5.14) app/controllers/apipie/apipies_controller.rb:28:in `index'
  actionpack (4.2.11) lib/action_controller/metal/implicit_render.rb:4:in `send_action'
  actionpack (4.2.11) lib/abstract_controller/base.rb:198:in `process_action'
  actionpack (4.2.11) lib/action_controller/metal/rendering.rb:10:in `process_action'
  actionpack (4.2.11) lib/abstract_controller/callbacks.rb:20:in `block in process_action'
[...]

PR #655, opened 2018-12-13
Fix authorization of resources

NielsKSchjoedt:
https://github.com/Apipie/apipie-rails/issues/654

### DIFF
--- a/app/controllers/apipie/apipies_controller.rb
+++ b/app/controllers/apipie/apipies_controller.rb
@@ -115,18 +115,28 @@ module Apipie
 
       new_doc = { :docs => @doc[:docs].clone }
 
-      new_doc[:docs][:resources] = @doc[:docs][:resources].select do |k, v|
-        if instance_exec(k, nil, v, &Apipie.configuration.authorize)
-          v[:methods] = v[:methods].select do |h|
-            instance_exec(k, h[:name], h, &Apipie.configuration.authorize)
-          end
-          true
-        else
-          false
+      new_doc[:docs][:resources] = if @doc[:docs][:resources].kind_of?(Array)
+        @doc[:docs][:resources].select do |resource|
+          authorize_resource(resource)
+        end
+      else
+        @doc[:docs][:resources].select do |_resource_name, resource|
+          authorize_resource(resource)
         end
       end
 
       new_doc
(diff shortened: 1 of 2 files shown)


lib/apipie/resource_description.rb
```diff
@@ -106,6 +106,7 @@ def to_json(method_name = nil, lang = nil)
 
       {
         :doc_url => doc_url,
+        :resource_name => _id,
```
iNecas: I wonder if we shouldn't call it `:id`, as we already have the `:name` field, and it might be misleading

iNecas:
Looks good, one comment to resolve

lib/apipie/resource_description.rb
NielsKSchjoedt:
```suggestion
        :id => _id,
```

app/controllers/apipie/apipies_controller.rb
```diff
@@ -115,20 +115,30 @@ def authorized_doc
...
+        @doc[:docs][:resources].select do |_resource_name, resource|
+          authorize_resource(resource)
         end
       end
 
       new_doc
     end
 
+    def authorize_resource resource
+      if instance_exec(resource[:resource_name], nil, resource, &Apipie.configuration.authorize)
```
NielsKSchjoedt:
```suggestion
      if instance_exec(resource[:id], nil, resource, &Apipie.configuration.authorize)
```

app/controllers/apipie/apipies_controller.rb
```diff
@@ -115,20 +115,30 @@ def authorized_doc
...
+        resource[:methods] = resource[:methods].select do |m|
+          instance_exec(resource[:resource_name], m[:name], m, &Apipie.configuration.authorize)
```
NielsKSchjoedt:
```suggestion
          instance_exec(resource[:id], m[:name], m, &Apipie.configuration.authorize)
```

lib/apipie/resource_description.rb
NielsKSchjoedt: Done

iNecas:
apipie-rails-0.5.15 including this change has just been released

PR #657, opened 2018-12-24
Added the required param to the docs

gatorjuice:
I think the documentation could be more clear that this param is required. Thanks peeps, love the software.


iNecas:
Thanks @gatorjuice

PR #658, opened 2019-01-03
Update travis ruby versions

iNecas:


iNecas:
This fixes `The program 'bundle' is currently not installed. To run 'bundle' please ask your administrator to install the package 'bundler'`. Seems like the issue is the bundler 2.0 is not compatible with ruby < 2.3.0

iNecas:
Seems like everything is green

PR #661, opened 2019-02-25
Update validators documentation

dirkjonker:
The number validator was missing and the decimal and array validator descriptions were mixed up.


iNecas:
Thanks @dirkjonker

PR #664, opened 2019-05-20
Add rails6 matrix to jenkins

iNecas:
+ remove some older rails versions to reduce the test matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,17 +11,16 @@ rvm:
   - 2.5.3
   - 2.6.0
 gemfile:
-  - Gemfile.rails41
   - Gemfile.rails42
-  - Gemfile.rails50 # Min ruby 2.2.2
   - Gemfile.rails51 # Min ruby 2.2.2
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -11,17 +11,10 @@ rvm:
   - 2.5.3
   - 2.6.0
 gemfile:
-  - Gemfile.rails41
   - Gemfile.rails42
-  - Gemfile.rails50 # Min ruby 2.2.2
   - Gemfile.rails51 # Min ruby 2.2.2
+  - Gemfile.rails60 # Min ruby 2.2.2
```
dirkjonker: Rails 6 requires ruby >= 2.5.0

iNecas: Updated

iNecas:
All green: there are still the deprecations, but I prefer merging this now before we merge something that would break on rails 6. Thanks @dirkjonker for the rspec workaround.

PR #665, opened 2019-05-20
Load file directly instead of using Rails constant loading to do so

### LINKED ISSUE #347: Controller concerns
I have a controller concern as follows:

```
module Pagination
  extend ActiveSupport::Concern

  included do
    before_action :set_limit_and_offset
  end

  def set_limit_and_offset
    # code here
  end
end
```

But when I load up the documentation url I get:

```
Unable to autoload constant Concerns::Pagination, expected /Users/<truncated>/app/controllers/concerns/pagination.rb to define it
```

speckins:
- Since Rails constant loading looks in Rails load paths, it's possible
  for files outside of api_controllers_matcher to be loaded
- Files that do not define an expected constant will cause a NameError
  to be raised.  This was most commonly encountered with files in the
  concerns/ directory, which caused a NameError for a missing
  "Concerns::<name>" constant.

This make use of ActiveSupport::Dependencies::Loadable#require_dependency
to explicitly load the file and its constants in a way that cooperates
with Rails constant loading and unloading.

https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#require-dependency
https://api.rubyonrails.org/classes/ActiveSupport/Dependencies/Loadable.html#method-i-require_dependency

Fixes #347.

### DIFF
--- a/lib/apipie/application.rb
+++ b/lib/apipie/application.rb
@@ -434,8 +434,7 @@ module Apipie
     end
 
     def load_controller_from_file(controller_file)
-      controller_class_name = controller_file.gsub(/\A.*\/app\/controllers\//,"").gsub(/\.\w*\Z/,"").camelize
-      controller_class_name.constantize
+      require_dependency controller_file
     end
 
     def ignored?(controller, method = nil)


iNecas:
Thanks @speckins for looking into this: mind rebasing: I've just updated the travis matrix and some unrelated fixes that caused the tests to fail

speckins:
@iNecas Rebased.

iNecas:
apipie-0.5.16 released to rubygems. Thanks @speckins

PR #673, opened 2019-09-19
Display engine path when using api!

derekwheel:
I am running issues with apipie where the paths of routes documented with `api!` are not displayed if they are in my Rails application's engines. 
The following PR seems to resolve the issue: https://github.com/Apipie/apipie-rails/pull/373
Given that the PR is a little older and has conflicts, I have taken that branch and rebased on current master, as well as doing some small updates to bring it current.

I'm happy to make any other necessary changes.

Thanks!

### DIFF
--- a/lib/apipie/core_ext/route.rb
+++ b/lib/apipie/core_ext/route.rb
@@ -1,10 +1,3 @@
-begin
-  # Rails 4
-  ActionDispatch::Journey
-rescue NameError
-  # Rails 3
-  require 'journey'
-  Journey
-end::Route.class_eval do
+ActionDispatch::Journey::Route.class_eval do
   attr_accessor :base_url
 end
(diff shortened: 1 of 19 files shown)


lib/apipie/core_ext/route.rb
```diff
@@ -0,0 +1,3 @@
+ActionDispatch::Journey::Route.class_eval do
+  attr_accessor :base_url
```
iNecas: If we really need to extend some class we don't own, I would prefer using a module to add the additional code, so that it's easier to track where it's coming from.

iNecas:
Thanks @derekwheel for picking this up. One small comment in the code, but otherwise, I think it's on good path (pun intended:) to get in

lib/apipie/core_ext/route.rb
derekwheel: Feedback addressed!

iNecas:
Squashed to two commit and merged in 103d39538f879. Thanks @derekwheel  and @neilgupta

PR #690, opened 2020-07-27
Skip extra parameters while validating the keys.

joshio1:
Currently, `validate_key` can be used to check whether any API is getting any extra parameters. The default behavior of `validate_key` is to raise an error if any unwanted parameter is found.

However, in many scenarios, I'd like to just ignore those extra parameters (i.e. those parameters should get deleted from the params) such that my controller action finally gets exactly what is defined. 

Currently, in order to achieve this, we have to again define those parameters which are required by our API. For example. we would do something like `params.permit(..)`. However, it'll be really great if I don't have to re-define those parameters at two places. Instead, apipie itself just deletes or ignores the unwanted or extra parameters.

This pull request defines one more key called as `action_on_non_validated_keys` which controls the default behavior of `validate_keys`. 
If `action_on_non_validated_keys` is set to `:raise`, it will work as it currently does. i.e. raising an error. (default behavior).
If `action_on_non_validated_keys` is set to `skip`, it will ignore the parameter and log the param without raising an error. By ignoring, it will delete that param from the params hash which are passed to the controller action.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -966,6 +966,9 @@ validate_presence
 validate_key
   Check the received params to ensure they are defined in the API. (false by default)
 
+action_on_non_validated_keys
+  Either `:raise` or `:skip`. If `validate_key` fails, raise error or skip and log the key (`:raise` by default)
+
 process_params
   Process and extract the parameter defined from the params of the request
   to the api_params variable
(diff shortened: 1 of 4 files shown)


mathieujobin (approved):
This looks great and makes a lot of sense, I love there is a configuration option for it.

can you please rebase?

PR #691, opened 2020-07-31
Display params in a nested format also in the API doc HTML.

joshio1:
Currently, if the parameters are deeply nested, it is very difficult to understand and comprehend the structure of the request and response params when viewing the API doc in the HTML format. This pull request is an attempt to display these nested parameters in a better way so that we can understand its structure in one go.

**Current**:

image

**Proposed**:

image

Note that the current structure is still there in the HTML. I have just supplemented that with this nested HTML structure having an hyperlink to the "normal" params.
Also, I have added tooltips which display the description of the params.

The entire idea is to grasp the entire request / response schema in one go. Please let me know what you think.

### DIFF
--- a/app/views/apipie/apipies/_method_detail.erb
+++ b/app/views/apipie/apipies/_method_detail.erb
@@ -24,6 +24,7 @@
 
 <% unless method[:params].blank? %>
   <%= heading(t('apipie.params'), h_level) %>
+  <%= render(:partial => "params_simple", :locals => {:params => method[:params]}) %>
   <table class='table'>
     <thead>
       <tr>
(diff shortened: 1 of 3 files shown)


DavertMik:
Is there a way to get this merged?

mathieujobin:
@joshio1 @DavertMik I think this would require a settings, so its OFF by default and people can opt in, if that's what they prefer.
I don't see why we should impose this on everyone.

ofedoren:
Although I agree with @mathieujobin here, I'd say that at first we better give users an option to switch this off and have it switched off by default, but mention this in the release notes as well as in README, so users can be actually aware of that. I like the proposed way, but it also seems like an experiment. In the future we could switch to this by default if more and more people will find this more useful.

mathieujobin:
@PanosCodes What do you think of this?

Issue #455, opened 2016-04-26
Documentation not working when same name of controllers reside in scopes


vkeziah:
I have 2 controllers having the same name but in different scopes and have the different routes but documentation not considering both , its considering only one , please look at the following example

```
  scope module: :courier do 
    resources :devices, path: "/couriers/devices"
  end

  scope module: :user do  
    resources :devices, path: "/users/devices"
  end
```
## courier devices controller

```
  api :POST, '/couriers/devices', 'Create device for courier'
```
## user devices controller

```
  api :POST, '/users/devices', 'Create device for user'
```

on documentation I see only couriers part like the following 

```
  POST /api/couriers/devices"   Create device for courier
```

I dont see my users url documentation, Whats wrong in that ? is this gem not working for scopes?

javvarcar:
Same here. I have 2 routes:
POST '/parts/favorites' in  parts/favorites#create 
POST 'stories/favorites' in stories/favorites#create

, and apipie only shows one.

I found this closed issue from 2013: https://github.com/Apipie/apipie-rails/issues/104

Any idea?

BrahimDa:
+1

yr0:
I've found the solution - actually the library supports it, but you have to opt-in in this behavior (for whatever reason). See https://github.com/Apipie/apipie-rails#configuration-reference - the configuration parameter `namespaced_resources` needs to be set to true:

```
# config/initializers/apipie.rb
Apipie.configure do |config|
  # ...
  config.namespaced_resources = true
end
```

Even after you do this, the names of api resources will be ugly (`MyNamespace::DetailsController` will be named `My namespacedetails`). I think this will be addressed in the future.

javvarcar:
@yr0 it works ;)

esen:
Before the ugly names addressed, monkey patching can save our eyes: 

```
module Apipie
  class Application
    def get_resource_name(klass)
      if klass.class == String
        klass
      elsif @controller_to_resource_id.has_key?(klass)
        @controller_to_resource_id[klass]
      elsif Apipie.configuration.namespaced_resources? && klass.respond_to?(:controller_path)
        return nil if klass == ActionController::Base
        path = klass.controller_path
        path.gsub(version_prefix(klass), " -> ").gsub("/", "-")
      elsif klass.respond_to?(:controller_name)
        return nil if klass == ActionController::Base
        klass.controller_name
      else
        raise "Apipie: Can not resolve resource #{klass} name."
      end
    end
  end
[...]

PR #694, opened 2020-08-26
Add support for namespaced controllers with the same name (response validation)

JorinTielen:
When you have two controllers with the same name, the response validation will confuse the two due to use of `controller_name`.
This passes the controller class in instead so namespacing information can be retained.

example situation:
```
class FooController; end

module Bar
  class FooController; end
end
```

### DIFF
--- a/lib/apipie/apipie_module.rb
+++ b/lib/apipie/apipie_module.rb
@@ -18,10 +18,10 @@ module Apipie
     app.to_swagger_json(version, resource_name, method_name, lang, clear_warnings)
   end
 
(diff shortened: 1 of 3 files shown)


mr-dxdy:
https://github.com/Apipie/apipie-rails/issues/455#issuecomment-268555114

mathieujobin (changes requested):
This needs a test to show the functionality, and ideally a documentation update.

mathieujobin:
Thanks for this, it looks simple and smart, but I don't understand without the tests being updated.
https://github.com/Apipie/apipie-rails/pull/855 was merged instead

PR #710, opened 2021-05-05
Add support for scheme definition in Swagger docs

danleyden:
A number of client generators require that the scheme is set in the swagger doc so they know to connect via HTTP or HTTPS.

This PR  adds support for defining the schemes supported by the API service, with the default being `['https']` (secure by default)


mathieujobin (approved):
This works with #711 ?
Seems highly related, I wonder why they needed to be on different PR ?

danleyden:
> This works with #711 ? Seems highly related, I wonder why they needed to be on different PR ?

They are related in terms of being improvements to generated Swagger docs, but they are for unrelated, independent features within the Swagger docs and each PR brings different value to the user.

mathieujobin:
This is the next one I would consider including... 
I would need the conflict resolved via a rebase. and someone else saying this is desirable.

ofedoren:
@mathieujobin, well, since the project already has **some** support for the OpenAPI, I'd say let's try to continue improving that support, so more people can find apipie-rails useful :) But it's all up to you.

PR #723, opened 2021-07-26
Change "an" to "a"

Naokimi:
When a vowel makes a consonant sound (like `y` or `w`) we write `a` instead of `an`

See https://english.stackexchange.com/questions/105116/is-it-a-user-or-an-user for more info


ofedoren (approved):
Thanks :)

PR #711, opened 2021-05-06
Support defining security mechanisms for Swagger

danleyden:
Generated Swagger documentation allows the API to specify what
authentication mechanisms are available and which are required.

This adds support (through configuration) for the API to include that
information in the generated swagger docs.
https://swagger.io/docs/specification/2-0/authentication/

PR #724, opened 2021-08-04
Swagger generator response examples

pkwen:
Swagger generator now supports `example` as an optional arg to `param` method


sasharevzin:
If you could squash your commits, that would b great!

mathieujobin:
Is this different from #710 and #711 ?

mathieujobin:
Please reopen if you think this is still required

Issue #565, opened 2017-10-19
Oops!! Method hola.en not found for resource "CONTROLLER_NAME"


VictorIbarra:
i have a 

```
Apipie.configure do |config|
  config.app_name                = "AppName"
  config.api_base_url            = "/api"
  config.doc_base_url            = "/apipie"
  # where is your API defined?
  config.api_controllers_matcher = "#{Rails.root}/app/controllers/**/*.rb"
end
```
And in my controller

`api :GET, '/hola'
  param :id, :number
  def hola
    puts "hola"
  end
`

i have a route in routes.rb get "hola"
when init apipie html page work but when enter 
http://localhost:3000/apipie/1.0/articles/hola.en.html

get a this error

http://localhost:3000/apipie/1.0/articles/hola.en.html

api_controllers_matcher is fine in apipie.rb
`config.api_controllers_matcher = "#{Rails.root}/app/controllers/**/*.rb"`

i use Rails 4.2.6

yjchieng:
You can fix the problem with:

`config.translate = false`

It looks like recent version has some problem. It is adding default locale ".en" into the path and you get your "hola.en.html" which points to nowhere.

luisalima:
Thanks @yjchieng , this worked! Is apipie still being maintained?

belgoros:
One year passed, the error is still there and no updates at least in the corresponding `README` section.

iNecas:
Just a reminder that this is an open-source community project: if you want to see the enhancements, the best way is to send a PR :)

PR #726, opened 2021-09-23
Fix 'Method foo.en not found for resource bar' errors with no languages set

drcapulet:
The default configuration has `default_locale = 'en'` but `languages = []`, so the check below fails. There's probably a cleaner way, but this logic matches that in `lib/tasks/apipie.rake` that treats the total set of languages to be `languages + default_local`.

#565 is at least one of the bug reports related to this.

### DIFF
--- a/app/controllers/apipie/apipies_controller.rb
+++ b/app/controllers/apipie/apipies_controller.rb
@@ -100,7 +100,7 @@ module Apipie
       [:resource, :method, :version].each do |par|
         if params[par]
           splitted = params[par].split('.')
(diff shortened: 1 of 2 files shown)


sasharevzin (approved):
Looks good to me

sasharevzin:
Can we move forward with this PR?

drcapulet:
I don't have permissions to merge, do you?

sasharevzin:
I don’t :(
@ofedoren are you around?

mathieujobin:
I can include this in an upcoming release.
Can you rebase your branch so we see that all tests pass?

thanks

mathieujobin (approved):

Issue #701, opened 2021-01-01
Minimization rails dependency


mr-dxdy:
Hi! I uses in my project only actionpack, activesupport, actionview, railties. 

May be change dependency `apipie-rails` from 
`s.add_dependency "rails", ">= 4.1"` 
to 
`s.add_development_dependency "rails", ">= 4.1"` 

I found pull-request [reduced rails dependency to development only](https://github.com/Apipie/apipie-rails/pull/266/files). Why was dependency rollback? I don't want install on my server actioncable, actionrecord and etc. I have a limit on free space on the server.

themilkman:
I just created a PR regarding this that you might be interested to try :)

mathieujobin:
see #741 as well

PR #729, opened 2021-11-05
Replace Rails with ActionView, -Pack and -Support

themilkman:
This PR reduces the required and thus installed gems to the only required ones on apps using apipie-rails, as also mentioned in #701.
I am looking for feedback regarding this.
Tests are passing (at least with ruby 2.7.4).

Cheers, Erik

### DIFF
--- a/apipie-rails.gemspec
+++ b/apipie-rails.gemspec
@@ -16,7 +16,11 @@ Gem::Specification.new do |s|
   s.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
   s.require_paths = ["lib"]
 
-  s.add_dependency "rails", ">= 4.1"
+  s.add_dependency "actionview"
+  s.add_dependency "actionpack"
+  s.add_dependency "activesupport"
+
+  s.add_development_dependency "rails", ">= 4.1"
   s.add_development_dependency "rspec-rails", "~> 3.0"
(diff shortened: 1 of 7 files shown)


apipie-rails.gemspec
```diff
@@ -16,7 +16,11 @@ Gem::Specification.new do |s|
   s.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
   s.require_paths = ["lib"]
 
-  s.add_dependency "rails", ">= 4.1"
+  s.add_dependency "actionview"
+  s.add_dependency "actionpack"
```
mathieujobin, line 20: how did you figured `actionpack` and `actionview` was required?
in #741 I am only depending on `railties` and `activemodel`...

themilkman: Actually I am not sure anymore (it's been a bit of time), but:
`railties` actually has `actionpack` as dependency. Thus my dependency should be smaller in this regard, I think.
Not sure why I added `actionview`, thought I'd have search the source code/the commit where the old dependency was replaced by rails.
But I am open to any other solution, as long as they work - everything is better than including `rails` with its raising dependencies. The smaller the footprint, the better it is.

mathieujobin:
actionview is a dependency of actionpack

I merged #741 instead.

thanks for the hints.

PR #733, opened 2022-01-04
allow_blank does not work

cs0511:
Current code only checks when allow_blank is true, it doesn't check if allow_blank is false.

If an attribute allows String, but doesn't want to allow empty string, allow_blank doesn't work at all.


stevehanson:
This would be useful on my project as well. We have been having to use regex types instead of `String` in order to validate that the string is not blank.

mathieujobin:
I think I have been bitten by this once as well... 
but 3 tests are failing for rails 4.2 
unless we can fix them, it means we would remove support and cut a major version

PR #740, opened 2022-01-26
Deprecate Travis, run tests on GitHub Actions

mathieujobin:
- Keeping 4.2, 5.2, 6.0 and 6.1 as supported
- Removing legacy rubies

Tests are failing with [ruby 3.0](https://github.com/Apipie/apipie-rails/compare/249f0d5b7c2...mathieujobin:ruby3) (should be a different PR, but I don't think I'll invest the time soon)
someone else can take it.


mathieujobin:
@ofedoren mind reviewing this?

mathieujobin:
@ofedoren you think we could move forward with this and #741 ?

.travis.yml
```diff
@@ -1,41 +0,0 @@
...
-rvm:
-  - 2.2.10
-  - 2.3.8
-  - 2.4.5
-  - 2.5.3
-  - 2.6.5
-  - 2.7.0
-gemfile:
-  - Gemfile.rails42
-  - Gemfile.rails51 # Min ruby 2.2.2
```
mathieujobin, line 16: Note that 4.1 and 5.0 was not running on CI before this change..

gemfiles/Gemfile.rails42.lock
```diff
@@ -0,0 +1,160 @@
+PATH
```
ofedoren, line 1: Why do we need to have this lock file?

mathieujobin: Simply to force Github Actions to select Bundler 1.17.3
this is required for Rails 4.2 tests to run on Github Actions. or at least, its the best solution I found that I use for many gems.
we can delete once we decide to drop Rails 4.2 from the test suite.

ofedoren (approved):
Thanks a lot, @mathieujobin!

PR #741, opened 2022-01-26
Only depends on railties and activemodel

mathieujobin:
Most of rails isn't required, and this gem should not bring all of it into minimalists apps, such as an API applications

includes #740

### DIFF
--- a/apipie-rails.gemspec
+++ b/apipie-rails.gemspec
@@ -16,7 +16,9 @@ Gem::Specification.new do |s|
   s.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
   s.require_paths = ["lib"]
 
-  s.add_dependency "rails", ">= 4.1"
+  s.add_dependency "railties", ">= 4.1"
+  s.add_dependency "activemodel", ">= 4.1"
+  s.add_dependency "activerecord", ">= 4.1"
   s.add_development_dependency "rspec-rails", "~> 3.0"
   s.add_development_dependency "sqlite3"
   s.add_development_dependency "minitest"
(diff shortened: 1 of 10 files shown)


mathieujobin:
@budnik Could you take a look at this please?

themilkman:
@mathieujobin Regarding the required gems, you could take a look at: https://github.com/Apipie/apipie-rails/pull/729 
I also hope to reduce the dependencies.

mathieujobin:
thanks @themilkman this is great as well
now, maybe I would like this one merge first because I am setting up github actions.
but if we can remove more after, that sounds good to me.

themilkman:
I think `railties` should be replaced with  `actionpack` and `activesupport` in order to get rid of the other dependencies `railties` carries (in addition to those two). Not sure if `activemodel` is required (as you know my branch uses `actionview` which might not be required, too).
Anyway, as long as it works either solution would be better than the current state.

mathieujobin:
@themilkman you are right ! `activemodel` doesn't appear to be required

taking `railties` only removes `thor` but sure, why not...

ofedoren:
@mathieujobin don't we actually depend on `ActionView` as well? https://github.com/Apipie/apipie-rails/blob/85aa3b58ae25390811bc13847e020317aeed397c/lib/tasks/apipie.rake#L175

themilkman:
I also reached this point and thought it should be necessary. Not sure why it works here anyway.

ofedoren:
Well, I didn't get to test the deps dropping PR, I just remember that in rake tasks (which can be used in production) we use `ActionView` directly and AFAIK we don't have test for tasks or at least basic, but also not sure why didn't fail...

mathieujobin:
@ofedoren @themilkman `actionview` is a dependency of `actionpack`.
That's why we don't need it explicitely.

themilkman:
Ha, gotcha @mathieujobin, this time I was blind regarding the nested dependencies, great :)

PR #744, opened 2022-02-22
fix: Controller resource set before version

LuukvH:
The resource description is set before the controller versions are defined.
When the namespaced_resources option is enabled this will lead to incorrect resource naming.


mathieujobin:
can you explain ? is it possible to write a test that reproduce the error this is fixing ?

mathieujobin:
also please update your branch to be up to date, so the tests runs

mathieujobin:
Thank you for updating your branch... I would not mind merging, but I need to understand what this fixes?
Is there a test that could be written that reproduce the problem that current master has and this fixes?

mathieujobin (changes requested):
I need to understand what problem this changes fix... ideally a test should be added

PR #749, opened 2022-04-01
Update boolean handling of false

colinbruce:
Expand the logic around blank value checking.

If `@allow_blank` is false, make sure it's not a BooleanValidator before looking for `value.blank?` as `false` is a valid boolean response, but
```ruby
false.blank? # => true
```
Create a test to replicate what we are seeing in use, when passing the value of `false` to a `:boolean` parameter (as opposed to `"false"` which is parsed as a string)

### DIFF
--- a/lib/apipie/param_description.rb
+++ b/lib/apipie/param_description.rb
@@ -117,7 +117,7 @@ module Apipie
       return true if @allow_nil && value.nil?
       return true if @allow_blank && value.blank?
       value = normalized_value(value)
-      if (!@allow_nil && value.nil?) || (!@allow_blank && value.blank?) || !@validator.valid?(value)
+      if (!@allow_nil && value.nil?) || (!@allow_blank && (!@validator.is_a?(Validator::BooleanValidator) && value.blank?)) || !@validator.valid?(value)
         error = @validator.error
         error = ParamError.new(error) unless error.is_a? StandardError
         raise error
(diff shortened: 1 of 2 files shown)


colinbruce:
I couldn't find a collaboration guide, so I hope this is okay?
If there are changes needed for CI, release notes, etc feel free to shout

colinbruce:
Agreed - I was only fixing the issue I saw! 👍 
Some analysis of #733 and regression testing of that may be useful @mathieujobin ?

mathieujobin:
@colinbruce Thank you so much for this quick fix and added tests.
I didn't think of this case when reviewing the other PR.
After I get another :+1: I will merge and release as 0.7.1

mathieujobin (approved):

dbussink:
> @colinbruce Thank you so much for this quick fix and added tests.

@mathieujobin I don't think this fix is complete then though? See also the comment earlier in https://github.com/Apipie/apipie-rails/pull/749#issuecomment-1087618548. Would more changes be necessary here?

mathieujobin:
@dbussink this still appears to be a good fix to add, the remaining issue can be fixed as a separate PR.
we need a test that reproduce it.

dbussink:
> @dbussink this still appears to be a good fix to add, the remaining issue can be fixed as a separate PR.
> we need a test that reproduce it.

What I was trying to say is that the issue extends to anything where `.blank?` returns true, not only for booleans then. But I'll look at a separate change if that's required then. I'm mostly not sure what the intended behavior is here and if it really needs a code change or a change to the release notes to indicate the change.

mathieujobin:
I am thinking we should actually question ActiveSupport as to why false.blank? Is true?

I don't know of any other objects that would have such a problem.

The allow_blank option was defaulting to false but wasn't enforced. So thing #733 and this PR is good.

I think I will add more tests with more object types.

Issue #755, opened 2022-04-18
apidocs always shows korean on master


mathieujobin:
I was testing the bootstrap update (#708) with my projects and I have a problem with `master`. Since PR #480
I get the Korean locale that I recently merged no matter what... 
it does not make sense, since it's only one new locale file being added... I tried setting `default_locale = :en`

if no one here knows what is going on, I think I'll force push v0.7.1 unto master, effectively removing the merge

PR #757, opened 2022-04-18
[Bugfix] API Docs were showing up in Korean by default

jsantos:
Fixes https://github.com/Apipie/apipie-rails/issues/755

It seems that since the Korean translations were added, the first key on the YAML file was `en`, which was recently causing the problem.


mathieujobin (approved):
Thank you !!!

PR #762, opened 2022-06-03
Allow custom validators to opt out of allow_blank behavior

stevehanson:
My app has a custom boolean validator that looks like this:

```ruby
class BoolValidator < Apipie::Validator::BaseValidator
  def validate(value)
    value.in?([true, false])
  end

    new(param_description) if argument == :bool
  end

  def description
    "Must be a boolean."
  end
end
```

But, this validator is failing when a `false` value is passed via the API because `false.blank?` is true and `allow_blank: false` is the default behavior for field definitions. There is a hard-coded [carve-out for BooleanValidator](https://github.com/Apipie/apipie-rails/blob/master/lib/apipie/param_description.rb#L128) so that this does not occur, but there is no way to replicate the same behavior in custom validators.

This PR adds a method to `BaseValidator` called `ignore_allow_blank?` and updates the blank validation logic so that it no longer has a hard-coded check for `BooleanValidator` and now just checks the validator's value for `ignore_allow_blank?`. The logic for `BooleanValidator` is preserved, and custom validators can now also achieve the same behavior.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1385,6 +1385,15 @@ options
 block
   Block converted into Proc, use it as you desire. In this example nil.
 
+If your validator includes valid values that respond true to `.blank?`, you
+should also define:
+
+.. code:: ruby
+   def ignore_allow_blank?
+     true
+   end
+
+so that the validation does not fail for valid values.
 
 ============
  Versioning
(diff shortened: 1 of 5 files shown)


mathieujobin (approved):
Truly awesome ! Thank you for this excellent patch

mathieujobin:
released as v0.8.2

PR #763, opened 2022-06-03
Add contributing instructions to readme

stevehanson:
I had some difficulty getting this project running since there isn't a Gemfile in the project, and I didn't see any contributing instructions. This PR adds some simple instructions for getting the test suite running to the Readme.

I do wonder why the Gemfile was removed and if the proper solution would be to add it back to the project.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1872,6 +1872,20 @@ provided it uses Apipie as a backend.
 
 And if you write one on your own, don't hesitate to share it with us!
 
+====================
+ Contributing
+====================
+
+Install dependencies:
+
+.. code:: shell
+   > bundle install --gemfile gemfiles/Gemfile.rails61
+
+Then run the tests:
+
+.. code:: shell
+   > export BUNDLE_GEMFILE='gemfiles/Gemfile.rails61'; bundle exec rspec
+
 
 ====================
  Disqus Integration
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -1872,6 +1872,20 @@ provided it uses Apipie as a backend.
...
 And if you write one on your own, don't hesitate to share it with us!
 
+====================
+ Contributing
+====================
+
+Since this gem does not have a Gemfile, you need to specify it in your shell with:
+
+.. code:: shell
+   BUNDLE_GEMFILE='gemfiles/Gemfile.rails61'
```
mathieujobin, line 1882: Thank you for this. you're right, this is helpful for new people.
Maybe we should keep a default Gemfile, I recently removed it and honestly, I don't remember why it felt redundant.

but this is the correct way to specify the rails version you'd expect.

that said, these new code blocks don't show up when I try to view the file in rendered mode.
maybe we need to specify `bash` instead of `shell` ?

I'm not familiar with `.rst` format actually.

mathieujobin:
Thank you for this patch

README.rst
mathieujobin: it was a dead symlink after I moved the gemfiles 39af442f2995e4555fb568216c6d790d923ff06b

so I deleted it

PR #765, opened 2022-06-30
Fix readme code block

stevehanson:
The code block in the readme documenting the `ignore_allow_blank?` functionality for custom validators that was added in https://github.com/Apipie/apipie-rails/pull/762 was not displaying. This fixes the RST syntax issue so it displays.


mathieujobin:
so a whiteline was needed?
wow, thanks

PR #768, opened 2022-08-20
Update readme with error handling example

sosaucily:
I struggled for a little too long to figure out how to best catch with error from a param validation failure. It was causing my server to 500, even in production. Yuck city. Thought this example would be helpful. Thanks for the library!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1160,6 +1160,7 @@ Here is an example of how to rescue and process a +ParamMissing+ or
 +ParamInvalid+ error from within the ApplicationController.
 
 .. code:: ruby
+  
   class ApplicationController < ActionController::Base
 
     # ParamError is superclass of ParamMissing, ParamInvalid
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -1156,6 +1156,20 @@ is raised and can be rescued and processed. It contains a description
 of the parameter value expectations. Validations can be turned off
 in the configuration file.
 
+Here is an example of how to rescue and process a +ParamMissing+ or
++ParamInvalid+ error from within the ApplicationController.
+
+.. code:: ruby
+  class ApplicationController < ActionController::Base
```
mathieujobin, line 1163: this code isn't displayed then the README is rendered.
I think you are only missing a white line after `code:: ruby`
can you try adding one?

mathieujobin (commented):
Thanks for the update, this is useful. please address the one comment I left and I will merge

README.rst
sosaucily: Good catch. Fixed! Looks better now

sosaucily:
Good catch @mathieujobin Fixed now!

mathieujobin (approved):

PR #769, opened 2022-08-23
Adds expected_type to IntegerValidator example

MrLeebo:
Otherwise, the generated docs.json outputs the field as a `string` and not a `number`


mathieujobin:
Thank you

mathieujobin:
released as v0.8.2

Issue #371, opened 2015-06-14
param_group inside block: undefined method `apipie_concern?' for nil:NilClass


olegantonyan:
Define a param_group

```
def_param_group :photo_urls do
  param :url_full, String  
  param :url_thumb, String
end
```

Then use it

```
api! 'Get all reports'
param_group :photo_urls
def index
....
```

And this will work just fine.
But put it inside param do block

```
api! 'Get all reports'
param :photo do
  param_group :photo_urls
end
def index
....
```

And boom!

```
undefined method `apipie_concern?' for nil:NilClass

/home/oleg/.rbenv/versions/2.2.2/lib64/ruby/gems/2.2.0/bundler/gems/apipie-rails-becc75b5f026/lib/apipie/dsl_definition.rb:336:in `param_group'
/home/oleg/projects/10levels-rails/app/controllers/api/v2/reports_controller.rb:32:in `block (2 levels) in <class:ReportsController>'
/home/oleg/.rbenv/versions/2.2.2/lib64/ruby/gems/2.2.0/bundler/gems/apipie-rails-becc75b5f026/lib/apipie/validator.rb:285:in `instance_exec'
/home/oleg/.rbenv/versions/2.2.2/lib64/ruby/gems/2.2.0/bundler/gems/apipie-rails-becc75b5f026/lib/apipie/validator.rb:285:in `initialize'
/home/oleg/.rbenv/versions/2.2.2/lib64/ruby/gems/2.2.0/bundler/gems/apipie-rails-becc75b5f026/lib/apipie/validator.rb:278:in `new'
/home/oleg/.rbenv/versions/2.2.2/lib64/ruby/gems/2.2.0/bundler/gems/apipie-rails-becc75b5f026/lib/apipie/validator.rb:278:in `build'
/home/oleg/.rbenv/versions/2.2.2/lib64/ruby/gems/2.2.0/bundler/gems/apipie-rails-becc75b5f026/lib/apipie/validator.rb:24:in `block in find'
/home/oleg/.rbenv/versions/2.2.2/lib64/ruby/gems/2.2.0/bundler/gems/apipie-rails-becc75b5f026/lib/apipie/validator.rb:23:in `each'
/home/oleg/.rbenv/versions/2.2.2/lib64/ruby/gems/2.2.0/bundler/gems/apipie-rails-becc75b5f026/lib/apipie/validator.rb:23:in `find'
[...]

PR #774, opened 2022-08-31
Fix issue where scope is nil when a param_group is accessed nested within response

olliswe:
Copy of https://github.com/Apipie/apipie-rails/pull/693 with tests added

Fixes https://github.com/Apipie/apipie-rails/issues/371


olliswe:
@mathieujobin as you asked in https://github.com/Apipie/apipie-rails/pull/693 I've added tests which assert the fix is working :)

olliswe:
@mathieujobin any ideas on if/when this will be merged ? :) My company is looking to take apipie on board to document our APIs !

mathieujobin:
released as v0.8.2

PR #786, opened 2022-09-20
add two tests when using an array of value for boolean

mathieujobin:


mathieujobin:
could someone provide some help fixing these new tests?

PanosCodes:
They were failing because the `BooleanValidator` did not accept `[true, false]` as argument. 

For example
```ruby
Apipie::Validator::BooleanValidator.build(param_description, [true, false], nil, nil)
```

would return `nil`

This commit https://github.com/Apipie/apipie-rails/pull/848/commits/b484c282c304da28024583b0c02cfb8715a7dd34 fixes that. I remember us talking about it [here](https://github.com/Apipie/apipie-rails/pull/803/files#r1058058655)


wdyt about https://github.com/Apipie/apipie-rails/pull/848 ?

PR #803, opened 2022-12-27
Move Swagger types and warnings under `/generator` namespace

PanosCodes:
#### Why
I'm woking on PR to make generators pluggable and allow external generators to be used by apipie.
This PR is a first step to start moving the Swagger 2.0 generator under `/generator` namespace.

#### How
It introduces 
 - `Apipie::Generator::Swagger::TypeExtractor` that is responsible to convert an `Apipie::Validator` to a Swagger type
 - `Apipie::Generator::Swagger::Warning` & `Apipie::Generator::Swagger::WarningWriter` to handle conditional logic on when a warning should be logged or not.
 
It's just a refactor and does not add any new functionality for the end user and it should not introduce any breaking changes.


lib/apipie/generator/generator.rb
```diff
@@ -0,0 +1,2 @@
+module Apipie::Generator
+end
```
mathieujobin, line 2: these small files only defining the module is good for auto-loading in Rails, but does not add much value in a gems where everything is explicitly `require`d 

it does not hurt either, except for adding extra files and bytes to the project.
I'm kinda leaning toward removing... you got another opinion?

lib/apipie/generator/swagger/type_extractor.rb
```diff
@@ -0,0 +1,69 @@
...
+      Apipie::Generator::Swagger::WarningWriter.instance.warn(warnings[:boolean])
+    end
+
+    extract
+  end
+
+  private
+
+  def extract
+    expected_type = if string?
```
mathieujobin, line 40: I prefer preventing wide indent by adding a return after the assignation.
can you update to

```ruby
expected_type = 
  if string?
    :string
  elsif boolean?
    :boolean
  elsif enum?
    :enum
  else
    @validator.expected_type.to_sym
  end
```

mathieujobin (commented):
This is a really cool refactor, I need more time to review, but I intend to merge...

lib/apipie/generator/generator.rb
PanosCodes: I used them in order for the `::` notation to work. 

For example 
```ruby 
class Apipie::Generator::Swagger::Type
end
```

instead of 
```ruby
module Apipie
  module Generator
    module Swagger
      class Type
      end
    end
  end
end
```

I'm fine with either way

lib/apipie/generator/swagger/type_extractor.rb
PanosCodes: That's fine with me.

Should we add it as a Rubocop cop ?

lib/apipie/generator/generator.rb
mathieujobin: right, that make sense... 💯  lets keep them instead adding the indent 👍🏽

lib/apipie/generator/swagger/type_extractor.rb
```diff
@@ -37,15 +37,16 @@ def extract_with_warnings(warnings = {})
...
-                    end
+    expected_type =
+      if string?
+        :string
+      elsif boolean?
+        :boolean
+      elsif enum?
+        :enum
+      else
+        @validator.expected_type.to_sym
```
mathieujobin, line 48: now that I am reading this properly.. maybe this was the original code... but I wonder why the `if/elsif/elsif/else` when `@validator.expected_type.to_sym` probably already return the right value for `string`, `boolean`, and `enum` ?

PanosCodes: What was confusing to me as well.
There are a couple of issues we handle here 

`ParamDescription`'s validator can be 
 - `nil`
 - `Apipie::Validator::BaseValidator`
 -  `ResponseDescriptionAdapter::PropDesc::Validator`

Updated Yard comment dc88775d6adbbea6f038b83ae93abd2446207169

____

In case of nil we want to default to `string`

____

`Apipie::Validator::EnumValidator#expected_type` [returns `string`](https://github.com/PanosCodes/apipie-rails/blob/87584003dd7ca78821c79ccfa20c97520a201e40/lib/apipie/validator.rb#L80)
____

`ResponseDescriptionAdapter::PropDesc::Validator#expected_type` [can we whatever](https://github.com/PanosCodes/apipie-rails/blob/15584d7acc2bc7c477ef4403b6f0fa05a60c6732/lib/apipie/response_description_adapter.rb#L63) by if we give it values that are `[true, false]` we need to result into `boolean`.

PR #804, opened 2022-12-28
Change output folder to `spec/dummy/tmp/`

PanosCodes:
#### Why
`spec/dummy/` is not version controlled. 
Using a not tracked folder will avoid test generated files being indexed by an IDE or git.

#### How
Changes the output folder to `spec/dummy/tmp/`


mathieujobin (approved):
Makes sense!

Also, I love your simple Why/How format description. Will copy.

PR #807, opened 2023-01-02
Add ruby 3.2 to the build mix

mathieujobin:


mtasaka:
So ruby3.2 failure maybe just due to `Fixnum` removal and `File.exists?` removal on ruby3.2...

(As far as I've checked) the first `uninitialized constant Apipie::Validator::IntegerValidator::Fixnum` error seems to be bringing about the following so many errors: Replacing `Fixnum` with `Integer` clears out most of the errors with leaving some errors due to   `File.exists?` removal, which can be fixed by using  `File.exist?`.

Although Fedora does not execute all of the tests, fixing the above 2 issues makes tests pass executed with ruby3.2 on Fedora rpm:

https://koji.fedoraproject.org/koji/taskinfo?taskID=96188951

mathieujobin:
Thanks for the hints @mtasaka 

I had not dug the issues, this is very helpful. I could apply them quickly and get it to pass.

PR #808, opened 2023-01-04
Fix #446 Fixing full_description translation

### LINKED ISSUE #446: Can't include a translation in the "description" option for actions.
I am trying to pass in a translation in the `description` option for one of my actions and the translation comes up as just plaintext. I am aware that the documentation states that the `description` option _will be converted to HTML_, but is there any support for this feature to allow me to pass translations in action descriptions?

Currently, using `I18n.t` method to translate in the `description` option _sort of_ works. When I have `config.languages` set in the `config/intializers/apipie.rb` file, it places the links for those languages at the top right of the documentation, but when clicking on a different link, it seems to only translate  on every other click when switching languages, and they become mismatched

**config/initializers/apipie.rb**

```
Apipie.configure do |config|
  config.app_name                = "Coolbeans API"
  config.api_base_url            = "/"
  config.doc_base_url            = "/docs"
  config.languages               = ['en', 'es']
  config.default_locale          = 'en'
  config.locale = lambda { |loc| loc ? I18n.locale = loc : I18n.locale }
  config.translate = lambda do |str, loc|
    old_loc = I18n.locale
    I18n.locale = loc
    trans = I18n.t(str)
    I18n.locale = old_loc
    trans
  end
  # where is your API defined?
  config.api_controllers_matcher = "#{Rails.root}/app/controllers/**/*.rb"
end
```

**V1::Accounts Controller**

```
api :POST, 'v1/accounts', "Creates a new account."
  description "apipie.accounts.create.below_is_an_example"   # => fails to translate
  error :code => 422, :desc => "Unprocessable Entity"
  example Account.example_request
  param :account, Hash, "The hash that holds the provided account field names and data.", :required => true do
    param :first_name,   String, "The new account's first name.", :required => true
    param :last_name,    String, "The new account's last name.", :required => true
    param :account_type, ["vendor", "member", "master", "admin"], "The new account's account type.", :required => true
    param :password,     String, "The new account's password.", :required => true
    param :active,       [true, false], "The new account's active status.", :required => true
  end

  def create
    account = Account.new(account_params)

    if account.save
      render :json => account
    else
      render :json => account.errors, :status => :unprocessable_entity
    end
  end
```

Antronin:
Full description cannot be translated as key was converted to HTML before translation. I changed that in this PR.

### DIFF
--- a/lib/apipie/error_description.rb
+++ b/lib/apipie/error_description.rb
@@ -34,7 +34,7 @@ module Apipie
     def to_json(lang)
       {
         :code => code,
-        :description => Apipe.app.translate(description, lang),
+        :description => Apipie.app.translate(description, lang),
         :metadata => metadata
       }
     end
--- a/lib/apipie/error_description.rb
+++ b/lib/apipie/error_description.rb
@@ -31,10 +31,10 @@ module Apipie
       end
     end
 
-    def to_json
+    def to_json(lang)
       {
         :code => code,
-        :description => description,
+        :description => Apipe.app.translate(description, lang),
         :metadata => metadata
       }
     end
--- a/lib/apipie/method_description.rb
+++ b/lib/apipie/method_description.rb
@@ -167,7 +167,7 @@ module Apipie
         :apis => method_apis_to_json(lang),
         :formats => formats,
         :full_description => Apipie.app.translate(@full_description, lang),
-        :errors => errors.map(&:to_json),
+        :errors => errors.map{ |error| error.to_json(lang) }.flatten,
         :params => params_ordered.map{ |param| param.to_json(lang) }.flatten,
         :returns => @returns.map{ |return_item| return_item.to_json(lang) }.flatten,
         :examples => @examples,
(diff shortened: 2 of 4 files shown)


Antronin:
I also added Response description translation and error description translation to this PR.

lib/apipie/error_description.rb
```diff
@@ -31,10 +31,10 @@ def initialize(code_or_options, desc=nil, options={})
       end
     end
 
-    def to_json
+    def to_json(lang)
       {
         :code => code,
-        :description => description,
+        :description => Apipe.app.translate(description, lang),
```
mathieujobin, line 37: maybe this typo makes everything fails?

lib/apipie/error_description.rb
```diff
@@ -31,10 +31,10 @@ def initialize(code_or_options, desc=nil, options={})
       end
     end
 
-    def to_json
+    def to_json(lang)
```
mathieujobin, line 34: maybe the param should have a default? I wonder if this cannot break people's applications...

mathieujobin (approved):

PR #810, opened 2023-01-12
Move swagger param description creation under `generator/swagger/param_description`

PanosCodes:
## Why
I'm woking on PR to make generators pluggable and allow external generators to be used by apipie.
This PR refactors and moves the param description generation under `lib/apipie/generator/swagger/param_description/` namespace.

### Introduced `Apipie::Generator::Swagger::ParamDescription::Composite`

Nested `ParamDescritpion`s like
```ruby
param :resource_param, Hash, :desc => 'Param description for all methods' do
  param :ausername, String, :desc => "Username for login", :required => true
  param :apassword, String, :desc => "Password for login", :required => true
end
```
are converted to swagger by `Apipie::Generator::Swagger::ParamDescription::Composite` that recursively loops though it's children.

`json_schema_obj_from_params_array` and `json_schema_param_defs_from_params_array` replaced by `Apipie::Generator::Swagger::Composite`

____

### Introduced `Apipie::Generator::Swagger::ParamDescription::Builder`
`swagger_atomic_param` replaced by `Apipie::Generator::Swagger::ParamDescription::Builder` that provides an interface responsible for constructing the Swagger parameters.  
____

### ParamDescription Swagger parameters
Swagger parameters that have some king of logic like `description`, `in`, `name`, `type`, `items`, `enum`

live under their respected classes
- `Apipie::Generator::Swagger::ParamDescription::Description`
- `Apipie::Generator::Swagger::ParamDescription::In`
- `Apipie::Generator::Swagger::ParamDescription::Name`
- `Apipie::Generator::Swagger::ParamDescription::Type` (This one also handles `items` & `enum` attributes)

and get instantiated through the `Apipie::Generator::Swagger::ParamDescription::Builder` 's api.

### Missing attributes

Instead of [creating](https://github.com/Apipie/apipie-rails/blob/master/lib/apipie/swagger_generator.rb#L437) an `OpenStruct` the mimics `ParamDescription` we create a `ParamDescription` from  that is required and has in it's options the information that is created from path (`added_from_path`) and not from the DSL. (`ParamDescription.create_for_missing_param`)

## Final thoughts
Although this PR adds more lines than it removes, I think it makes it easer to navigate the generator code by having different classes for different responsibilities.


lib/apipie/swagger_generator.rb
```diff
@@ -434,105 +430,14 @@ def add_missing_params(method, path)
 
       missing.each do |name|
         warn_path_parameter_not_described(name, path)
-        result[name.to_sym] = OpenStruct.new({
```
mathieujobin, line 437: Always happy to see an OpenStruct being removed

mathieujobin (commented):
would be great to see an update to the README about the new Composite feature.

PanosCodes:
> would be great to see an update to the README about the new Composite feature.

It's meant to be an internal implementation for swagger generator, maybe once the refactor is done we could have a `README.md`  under `generator/swagger` giving some context 🤷‍♂️

PR #836, opened 2023-03-03
Fixed warnings due to uninitialized instance variable

nitinsingh25:
I have fixed following two warnings:

1. lib/apipie/param_description.rb:103: warning: instance variable @from_concern not initialized
2. lib/apipie/dsl_definition.rb:349: warning: instance variable @_current_param_group not initialized

### DIFF
--- a/lib/apipie/param_description.rb
+++ b/lib/apipie/param_description.rb
@@ -101,7 +101,7 @@ module Apipie
     end
 
     def from_concern?
-      method_description.from_concern? || @from_concern
+      method_description.from_concern? || (defined?(@from_concern) && @from_concern)
     end
 
     def normalized_value(value)
--- a/lib/apipie/dsl_definition.rb
+++ b/lib/apipie/dsl_definition.rb
@@ -344,10 +344,11 @@ module Apipie
       #
       def param(param_name, validator, desc_or_options = nil, options = {}, &block) #:doc:
         return unless Apipie.active_dsl?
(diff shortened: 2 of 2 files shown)


lib/apipie/param_description.rb
```diff
@@ -101,7 +101,7 @@ def initialize(method_description, name, validator, desc_or_options = nil, optio
     end
 
     def from_concern?
-      method_description.from_concern? || @from_concern
+      method_description.from_concern? || (defined?(@from_concern) && @from_concern)
```
PanosCodes, line 104: I think here we can update lines

https://github.com/Apipie/apipie-rails/blob/b0bc6d60dbe2cedfa155ba4939f71b1943766509/lib/apipie/param_description.rb#L56-L58

from 
```ruby
if @options[:param_group]
  @from_concern = @options[:param_group][:from_concern]
end
```

to 
```ruby
@from_concern = @options.dig(:param_group, :from_concern)
```

mathieujobin (changes requested):
I don't think adding a check on usage is the right way. just making sure it gets initialize on object creation is probably more desirable

Thanks for taking a stab at it though.

What do you think of https://github.com/Apipie/apipie-rails/pull/838 ?

nitinsingh25:
> I don't think adding a check on usage is the right way. just making sure it gets initialize on object creation is probably more desirable
> 
> Thanks for taking a stab at it though.
> 
> What do you think of #838 ?

Yup, changes of #838 seems more appropriate. Let me test my changes with #838.

PR #840, opened 2023-03-08
Disables FrozenStringLiteralComment rule

PanosCodes:
### Why
It does not seem to make a lot of sense to be enforced anymore ([Standard](https://github.com/testdouble/standard#why-arent-frozen_string_literal-true-magic-comments-enforced) has a good explanation) and since it's not used in the gem's codebase I think it can be disabled.

### DIFF
--- a/lib/apipie/errors.rb
+++ b/lib/apipie/errors.rb
@@ -60,27 +60,13 @@ module Apipie
 
   class ResponseDoesNotMatchSwaggerSchema < Error
     def initialize(controller_name, method_name, response_code, error_messages, schema, returned_object)
-      @controller_name = controller_name
-      @method_name = method_name
-      @response_code = response_code
-      @error_messages = error_messages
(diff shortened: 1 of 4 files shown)


lib/apipie/errors.rb
```diff
@@ -60,27 +60,13 @@ def to_s
...
 
   class NoDocumentedMethod < Error
     def initialize(controller_name, method_name)
-      @method_name = method_name
-      @controller_name = controller_name
-    end
-
-    def to_s
-      "There is no documented method #{@controller_name}##{@method_name}"
+      super("There is no documented method #{controller_name}##{method_name}")
```
mathieujobin, line 69: this is really cool, I'm going to reuse this pattern.

mathieujobin (approved):

PanosCodes:
Can we merge this ? 😅

PR #841, opened 2023-03-08
Disables Style/IfUnlessModifier rule

PanosCodes:
Having this rule enabled forces the modifier if statement to be used

So this

```ruby
if Apipie.configuration.swagger_api_host.present?
  @swagger[:host] = Apipie.configuration.swagger_api_host
end
```

should be changed to 

```ruby
@swagger[:host] = Apipie.configuration.swagger_api_host if Apipie.configuration.swagger_api_host.present?
```

resulting in longer, harder to read lines.

Since the use the regular if statement all over the gem's codebase we can disable this rule.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -55,6 +55,9 @@ Naming/PredicateName:
   AllowedMethods:
     - have_field?
 
+Style/IfUnlessModifier:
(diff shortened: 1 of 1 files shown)


mathieujobin:
Alright

PR #816, opened 2023-02-07
Refactor Swagger generator

PanosCodes:
## Why
In the spirit of https://github.com/Apipie/apipie-rails/pull/803 and https://github.com/Apipie/apipie-rails/pull/810 this PR is an attempt to break down the Swagger generator into smaller modules.

## How

The generator has only the three entrypoint class methods(`generate_from_resources`, `json_schema_for_method_response` & `json_schema_for_self_describing_class` and the rest of the code has been moved to smaller files.

## Thoughts
Again, although this PR adds more lines than it removes, I think it makes it easer to navigate and test the generator code.

PR #843, opened 2023-03-10
Disables Metrics/ParameterLists

PanosCodes:
Although long parameters can be considered a code smell, in case of a [data transfer objects ](https://github.com/Apipie/apipie-rails/pull/816/files#diff-3926cca4c7fc6e45d5a46a2a0d7c9b20172fc231774be32a73cd362a1d81d231R5) makes sense if the parameters are more than 7

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -51,6 +51,9 @@ Metrics/BlockLength:
     - lib/tasks/apipie.rake
     - lib/apipie/swagger_generator.rb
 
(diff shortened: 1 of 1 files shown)


mathieujobin:
So the default is 7 ?
maybe it does not hurt to maintain an ignore list for the few exceptions?

I actually think 7 params is quite high for a method.

PanosCodes:
Sure

PR #848, opened 2023-03-20
Allow boolean array `[true, false]` to be a valid argument for BooleanValidator

PanosCodes:
### How

- Allow BooleanValidator to be build using `[true, false]` as argument.
- Remove the `[true, false]` handling from Swagger generator

____

It would probably be easier to review per commit


mjobin-mdsol (commented):
Looks good, I'm slightly afraid this could introduce a breaking change... ?

PanosCodes:
We can always put it behing and setting for example
```ruby
Apipie.configure do |config|
  config.validator.boolean.handle_boolean_array = false
end
```
My question is why use `[true, false]` in the first place instead of `:bool` or `:boolean` ?
If we deside to allow `[true, false]` passed as validator should we also allow `[1, 0]`

I'can't think any compelling reason and I'm leaning in closing this PR 😅 (we can use the tests refactoring in another PR)

mathieujobin:
I don't know, we can keep it open for a while and wait if others wants to chime in their opinion.

mathieujobin (approved):
if you want to solve the conflict after the refactor, I will merge this as well.

mathieujobin:
Thanks, I wasn't exactly sure about the spec removal...

PR #866, opened 2023-04-25
Update README.rst

daprezjer:
Adds the new Integer option you added recently to the documentation here.

Any chance I could get maintainer access? I'd like to suggest the other types and potentially other improvements, without having to go through the whole forking process.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1205,7 +1205,7 @@ after the ``apipie_validations`` before_action.
 
 TypeValidator
 -------------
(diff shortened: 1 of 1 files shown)


mathieujobin (approved):

mathieujobin:
@daprezjer Thanks for your contribution.
Can I ask you to keep using your fork for a while? Other contributors can be very active without write access to the main repo. Let me know what overhead it causes you, because it should be almost the same.

PR #869, opened 2023-05-11
Fix startup crash due to typo

stormsilver:
This should be `get_resource_id`. The method `remove_method_description` was not exercised by the specs so it wasn't caught. (see #864)

This was causing apipie to crash during application boot.


PanosCodes (approved):
LGTM 🚀, Thanks for opening a PR

PR #870, opened 2023-05-11
Improve performance of route detection

stormsilver:
In large apps with many routes (1500+), using the `api!` method would be quite slow because `routes_for_action` looped over every route trying to constantize the controller name. Compound this by the number of times `api!` is used in the app.

Furthermore, each controller utilizing `api!` would constantize the next controller using `api!`, which would constantize the next controller and so on, recursively and potentially very deeply.

This change caches a hash of controller+action => routes using only the controller name, avoiding the time-consuming loops, constantization, and recursion.

As a bonus, the cache is cleared when documentation is reloaded, fixing a bug where new routes added while the app is running wouldn't get picked up.

On small apps with not too many routes, there isn't much of a performance gain.

To test this, I added 1000 routes to the dummy app's routes file.

## Before
image

## After
image

## Master
(for comparison - without 1000 routes)
image

### DIFF
--- a/lib/apipie/application.rb
+++ b/lib/apipie/application.rb
@@ -26,14 +26,13 @@ module Apipie
     end
 
     def rails_routes(route_set = nil, base_url = "")
-      if route_set.nil? && @rails_routes
-        return @rails_routes
-      end
+      return @_rails_routes if route_set.nil? && @_rails_routes
+
       route_set ||= Rails.application.routes
       # ensure routes are loaded
       Rails.application.reload_routes! unless Rails.application.routes.routes.any?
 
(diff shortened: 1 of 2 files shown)


drexed (approved):

mathieujobin:
Wonderful, thanks a lot

patrickshine:
@mathieujobin thank you for the quick attention on the recent PRs! Would you be able to publish a patch/minor version so we can pull the latest version into our application? Thanks!

mathieujobin:
You got it, v1.1.0 released

patrickshine:
Amazing, thank you again!

PR #871, opened 2023-05-11
Skip parse body for pdf responses

drexed:
PDF responses return binary data that cannot be parsed correctly.


mathieujobin:
Makes sense, thanks

PR #872, opened 2023-05-15
Allow resource_name to be inherited

stormsilver:
This is helpful if you have a very structured hierarchy of API controllers and would like to define the naming logic once in a base controller.

* Allow resource_name to be a string, symbol, or proc
* Search all the way up the controller hierarchy for a value
* Continue searching even if a particular controller doesn't have a `resource_description` block

Edit: I had to set Rubocop's target Ruby version to 2.6 in order to get a bunch of offenses to go away. I'm not sure why that suddenly started happening (my PR from last week didn't have any issues).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,7 @@ Rails:
 
 AllCops:
   NewCops: enable
+  TargetRubyVersion: 2.6
 
 Layout/LineLength:
   Max: 140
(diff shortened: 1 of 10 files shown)


PanosCodes:
> define the naming logic once in a base controller.

image

mathieujobin (approved):
This looks all great!!! Thank you

PR #874, opened 2023-05-22
Fix cache rendering with namespaced resources

stormsilver:
The path traversal sanitization was incorrectly filtering out paths that included the `-` character. Namespaced resources generates files like `api-v1-resource.html` which were then returning a 400.

### DIFF
--- a/app/controllers/apipie/apipies_controller.rb
+++ b/app/controllers/apipie/apipies_controller.rb
@@ -154,26 +154,12 @@ module Apipie
 
     def render_from_cache
       path = Apipie.configuration.doc_base_url.dup
-      # some params can contain dot, but only one in row
-      if [:resource, :method, :format, :version].any? { |p| params[p].to_s.gsub(".", "") =~ /\W/ || params[p].to_s.include?('..') }
-        head :bad_request and return
-      end
-
       path << "/" << params[:version] if params[:version].present?
       path << "/" << params[:resource] if params[:resource].present?
       path << "/" << params[:method] if params[:method].present?
-      if params[:format].present?
-        path << ".#{params[:format]}"
-      else
-        path << ".html"
-      end
-
-      # we sanitize the params before so in ideal case, this condition
-      # will be never satisfied. It's here for cases somebody adds new
-      # param into the path later and forgets about sanitation.
-      if path.include?('..')
-        head :bad_request and return
-      end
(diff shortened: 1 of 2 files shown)


juan-belle (approved):

app/controllers/apipie/apipies_controller.rb
```diff
@@ -154,26 +154,12 @@ def get_format
...
-      # some params can contain dot, but only one in row
-      if [:resource, :method, :format, :version].any? { |p| params[p].to_s.gsub(".", "") =~ /\W/ || params[p].to_s.include?('..') }
-        head :bad_request and return
-      end
-
       path << "/" << params[:version] if params[:version].present?
       path << "/" << params[:resource] if params[:resource].present?
       path << "/" << params[:method] if params[:method].present?
-      if params[:format].present?
-        path << ".#{params[:format]}"
```
mathieujobin, line 166: but we are not adding this anymore?

app/controllers/apipie/apipies_controller.rb
```diff
@@ -154,26 +154,12 @@ def get_format
...
-
-      # we sanitize the params before so in ideal case, this condition
-      # will be never satisfied. It's here for cases somebody adds new
-      # param into the path later and forgets about sanitation.
-      if path.include?('..')
-        head :bad_request and return
-      end
+      # Sanitize path against directory traversal attacks (e.g. ../../foo)
+      # by turning path into an absolute path before appending it to the cache dir
+      path = File.expand_path("#{path}.#{request.format.symbol}", '/')
```
mathieujobin, line 162: I see you are taking the format from the request, but can it be different?

stormsilver: Hi! Sorry I didn't respond earlier - yesterday was really busy.

Since Rails/Rack assemble the request format for us, I figured it was safest to use `request.format` rather than us needing to sanitize `params[:format]` ourselves. It's one small security item we don't have to do.

If we had a large variety request formats it might be possible for `request.format.symbol` to be different from the file extension on the file system, but we only have `html` and `json` which are pretty standard. I don't think it can be different.

I'm more than willing to change this in a new PR if you think we should. Thank you for merging it! (And all my other recent PRs - you are super responsive which is not often found 🎉 )

PR #879, opened 2023-05-26
Fix deprecated `content_type` on Rails >= 6

stormsilver:
DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. (called from analyze_functional_test at ~/apipie-rails/lib/apipie/extractor/recorder.rb:51)


mathieujobin:
Thanks

PR #881, opened 2023-05-30
Swagger: Adds option to skip default tags

### LINKED ISSUE #759: [Feature request] Remove default tags from generated swagger json
Hello
When using swagger ui with the swagger json generated by apipie. The jsons always contain a default tag (the controller name or path) even if I specify tags in the resource_description, or give the resource a unique name, this results in multiple occurrences of the same endpoint on the swagger ui. Can I somehow get rid of the tag that was not given by me?
Thank you!

If I read the code properly, currently it is always there currently https://github.com/Apipie/apipie-rails/blob/master/lib/apipie/swagger_generator.rb#L229

davidwessman:
- This option allows the resource tags to be manually specified
- Fixes #759

### DIFF
--- a/lib/apipie/generator/swagger/method_description/api_schema_service.rb
+++ b/lib/apipie/generator/swagger/method_description/api_schema_service.rb
@@ -47,9 +47,12 @@ class Apipie::Generator::Swagger::MethodDescription::ApiSchemaService
   end
 
   def tags
-    [@method_description.resource._id] +
-      warning_tags +
-      @method_description.tag_list.tags
+    tags = if Apipie.configuration.generator.swagger.skip_default_tags?
+      []
+    else
+      [@method_description.resource._id]
+    end
+    tags + warning_tags + @method_description.tag_list.tags
   end
 
   def warning_tags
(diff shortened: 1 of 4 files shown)


lib/apipie/generator/swagger/method_description/api_schema_service.rb
```diff
@@ -47,9 +47,12 @@ def summary(api)
   end
 
   def tags
-    [@method_description.resource._id] +
-      warning_tags +
-      @method_description.tag_list.tags
+    tags = if Apipie.configuration.generator.swagger.skip_default_tags?
```
davidwessman, line 50: Maybe it should be called `skip_default_resource_tag` since it still allows warning_tags? (Which has a separate config)

mathieujobin:
@PanosCodes can you please review this Swagger related change?

lib/apipie/generator/swagger/method_description/api_schema_service.rb
PanosCodes: If I'm not mistaken swagger tags are only used in endpoints so I guess it's ok to leave it as is, fine by me either way.

PanosCodes (approved):
Nice work. LGTM 😀 

One thing to keep in mind is that when no tags exist the expected SwaggerUI behavior is to group them under the `default` section. image

Feel free to check the Rubocop violations.

davidwessman:
@PanosCodes Fixed the rubocop warnings and rebased 🙂 
@mathieujobin Good to go? 🙂

PR #886, opened 2023-06-08
Parameter validation: Raises error for all missing

### LINKED ISSUE #802: Report all missing required parameters instead of just the first
In the case of an endpoint having multiple required params, would it be possible to include all missing params in the error message? As it is, it looks like these are reported one-by-one in order:

https://github.com/Apipie/apipie-rails/blob/6a0174d20d3ccfe9a0a528826eb4abe886728021/lib/apipie/dsl_definition.rb#L246-L251

davidwessman:
- Instead of just raising `ParamMissing` for the first missing,
  instead raise a compound `ParamMultipleMissing` if there are more
  than one missing.
- Adds specs for both POST and GET requests.
- Fixes #802


lib/apipie/errors.rb
```diff
@@ -24,6 +24,20 @@ def initialize(param)
     end
   end
 
+  class ParamMultipleMissing < ParamError
```
davidwessman, line 27: Would be nice to inherit from the same error as `ParamMissing`, but we do not have the `defined param` here.

davidwessman:
I have no good way to solve the last rubocop warnings, do you usually add ignores if the complexity grows? Not sure if I can split it up more since it is meta-programming

PanosCodes:
Running `rubocop --auto-gen-config --exclude-limit 180` will add the errors to `.rubocop_todo.yml`

mathieujobin (approved):
Yep, its OK to ignore `ModuleLength` and `BlockLength` rubocop offenses, you can use the comment as @PanosCodes suggested. and I will merge.

Thank you

mathieujobin (approved):

davidwessman:
Thanks! 🙂

Issue #888, opened 2023-06-13
Swagger warnings for optional parameters even when using `property`


davidwessman:
Hello!

When I specify my return values like this:
```ruby
          api(
            :POST,
            "/v2/namespace/credentials",
            "Generate new credentials"
          )
          returns(code: 200) do
            property(:username, String)
            property(:password, String)
          end
          def create
          end
```

and generate swagger, I still get warnings:
```ruby
The parameter :username is optional but default value is not specified (use :default_value => ...)
The parameter :password is optional but default value is not specified (use :default_value => ...)
```

even though `property` are required by default. I also tried adding `required: true` but still get the errors.

It does generate the correct swagger file as far as I understand:
```json
"responses": {
          "200": {
            "description": "OK",
            "schema": {
              "type": "object",
              "properties": {
                "username": {
                  "type": "string"
                },
                "password": {
                  "type": "string"
                }
              },
              "additionalProperties": false,
              "required": [
                "username",
                "password"
              ]
            }
          }
        },
```

davidwessman:
The `@in_schema` seems to be true and then it wont consider `@param_description.required` which would have returned true:
https://github.com/Apipie/apipie-rails/blob/61ae61d1247543dcb17ebb58267dc5ca0e987c7f/lib/apipie/generator/swagger/param_description/builder.rb#L82

davidwessman:
Fixed in 1.2.2, thank you @shev-vadim-net 🎉

PR #892, opened 2023-07-03
Fix Swagger warnings for properties

shev-vadim-net:
This PR should fix the #888 issue. I'm not sure how exactly the `in_schema` is used and whether we can just ignore it in `Apipie::Generator::Swagger::ParamDescription::Builder`. But since the `in_schema` is almost always `true`, I couldn't find a way how to combine it with the `param_description` in other way.

### DIFF
--- a/lib/apipie/generator/swagger/param_description/builder.rb
+++ b/lib/apipie/generator/swagger/param_description/builder.rb
@@ -79,7 +79,7 @@ class Apipie::Generator::Swagger::ParamDescription::Builder
   end
 
   def required?
-    required_from_path? || (!@in_schema && @param_description.required)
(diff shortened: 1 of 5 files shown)


lib/apipie/generator/swagger/param_description/builder.rb
```diff
@@ -79,7 +79,7 @@ def for_default
   end
 
   def required?
-    required_from_path? || (!@in_schema && @param_description.required)
+    required_from_path? || @param_description.required
```
mathieujobin, line 82: @PanosCodes can you confirm we can remove this check here?

mathieujobin:
Can you run the `rubocop --auto-correct` and push the changes ?
thanks

PR #894, opened 2023-07-31
param: Consider default_value: nil as valid config

davidwessman:
When using `param` as:
```ruby
param(:name, String, required: false, default_value: nil)
```

there would be a warning that `default_value` is missing.
This fix changes from checking `options[key].blank?` to `options.key?(key)`.

### DIFF
--- a/lib/apipie/generator/swagger/param_description/builder.rb
+++ b/lib/apipie/generator/swagger/param_description/builder.rb
@@ -71,7 +71,7 @@ class Apipie::Generator::Swagger::ParamDescription::Builder
   end
 
   def for_default
-    return {} if @param_description.options[:default_value].blank?
+    return {} unless @param_description.options.key?(:default_value)
 
     {
       default: @param_description.options[:default_value],
--- a/spec/lib/apipie/generator/swagger/param_description/builder_spec.rb
+++ b/spec/lib/apipie/generator/swagger/param_description/builder_spec.rb
@@ -102,7 +102,7 @@ describe Apipie::Generator::Swagger::ParamDescription::Builder do
 
     subject { generated_block }
 
-    context 'required: true' do
+    context 'when required is true' do
       let(:base_param_description_options) { { required: true } }
 
       it 'does not output an option without default warning' do
@@ -112,8 +112,8 @@ describe Apipie::Generator::Swagger::ParamDescription::Builder do
       end
     end
 
-    context 'required: false' do
-      context 'no warning if default_value: nil' do
+    context 'when required is false' do
+      context 'when default_value is nil' do
         let(:base_param_description_options) do
           { required: false, default_value: nil }
         end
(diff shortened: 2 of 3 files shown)


spec/lib/apipie/generator/swagger/param_description/builder_spec.rb
```diff
@@ -102,23 +102,49 @@
...
+          expect { subject }.not_to output(
+            /is optional but default value is not specified/
+          ).to_stderr
         end
       end
-    end
 
-    context 'when is required' do
-      let(:base_param_description_options) { { required: true } }
+      context 'no warning if default_value: 123' do
```
PanosCodes, line 128: NIT We can use the `context` to describe and the environment and let the `it` hint what the result should be  
```suggestion
      context 'when default_value is 123' do
```

spec/lib/apipie/generator/swagger/param_description/builder_spec.rb
```diff
@@ -102,23 +102,49 @@
 
     subject { generated_block }
 
-    context 'when is not required' do
-      let(:base_param_description_options) { { required: false } }
+    context 'required: true' do
```
PanosCodes, line 105: NIT 
```suggestion
    context 'when required is true' do
```

spec/lib/apipie/generator/swagger/param_description/builder_spec.rb
```diff
@@ -102,23 +102,49 @@
...
+      end
+    end
 
-      context 'and no default is given' do
-        before { param_description_options.delete(:default) }
+    context 'required: false' do
+      context 'no warning if default_value: nil' do
+        let(:base_param_description_options) do
+          { required: false, default_value: nil }
+        end
```
PanosCodes, line 119:
```suggestion
    context 'when required is false' do
      let(:base_param_description_options) { { required: false } }
      
      context 'when default_value: nil' do
        let(:base_param_description_options) do
          { required: false, default_value: nil }
        end
```

PanosCodes (approved):
LGTM 
I'm not really sure about combining it with `allow_nil` & `allow_blank`🤷‍♂️

lib/apipie/generator/swagger/param_description/builder.rb
```diff
@@ -71,7 +71,7 @@ def for_required
   end
 
   def for_default
-    return {} if @param_description.options[:default_value].blank?
```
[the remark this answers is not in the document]
mathieujobin: No, I don't think they should be combined.

mathieujobin (approved):
Thanks for the fix

davidwessman:
@mathieujobin Could we get a release of this? 🙂

mathieujobin:
sure, sorry for the delay on this. will be in v1.2.3 shortly

davidwessman:
> sure, sorry for the delay on this. will be in v1.2.3 shortly

No worries, just a small fix.

Thank you 👏🏻

PR #901, opened 2023-10-28
support property example

kuei0221:
[WHY]
* The generated swagger JSON didn't include the example for the response. This will be useful when the user wants to know the response example, this will help them to understand the API quicker.

[HOW]
* add a new option for the params description `ex` to provide an example for the specific parameters

Here is a swagger editor preview with the ex option, which correctly shows the example in response:
image

### DIFF
--- a/lib/apipie/generator/swagger/param_description/builder.rb
+++ b/lib/apipie/generator/swagger/param_description/builder.rb
@@ -53,6 +53,7 @@ class Apipie::Generator::Swagger::ParamDescription::Builder
 
     definition.merge!(for_default)
     definition.merge!(for_required)
+    definition.merge!(for_example)
     definition.merge!(@description.to_hash) if @description.present?
 
     warn_optional_without_default_value(definition)
@@ -75,6 +76,14 @@ class Apipie::Generator::Swagger::ParamDescription::Builder
 
     {
       default: @param_description.options[:default_value],
+    }
+  end
+
+  def for_example
+    return {} unless @param_description.options.key?(:ex)
+
+    {
+      example: @param_description.options[:ex],
     }
   end
 
--- a/lib/apipie/generator/swagger/param_description/builder.rb
+++ b/lib/apipie/generator/swagger/param_description/builder.rb
@@ -83,7 +83,7 @@ class Apipie::Generator::Swagger::ParamDescription::Builder
     return {} unless @param_description.options.key?(:ex)
 
     {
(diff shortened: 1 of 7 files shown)


lib/apipie/generator/swagger/param_description/builder.rb
```diff
@@ -78,6 +79,14 @@ def for_default
     }
   end
 
+  def for_example
+    return {} unless @param_description.options.key?(:ex)
+
+    {
+      example: @param_description.options[:ex],
```
PanosCodes, line 86:
```suggestion
      example: @param_description.options[:example],
```

spec/lib/apipie/generator/swagger/param_description/builder_spec.rb
```diff
@@ -200,4 +200,16 @@
...
+
+  describe 'example' do
+    subject { generated_block[:example] }
+
+    it { is_expected.to be_blank }
+
+    context 'when example is assigned' do
+      let(:base_param_description_options) { { ex: 'example'} }
+
+      it { is_expected.to be_present }
```
PanosCodes, line 212:
```suggestion
      it { is_expected.to eq('example') }
```

Let's be explicit

PanosCodes (changes requested):
Awesome work 🚀 Thank you for working on it 🙏 `example` is indeed a [valid swagger attribute](https://swagger.io/docs/specification/2-0/adding-examples/).

Perhaps the options can be named `example`? Do you know if there was a reason not to do that?

lib/apipie/generator/swagger/param_description/builder.rb
kuei0221: I agree it will be better to name it explicitly 😃. My only concern is that there is already an `example` keywords in the method description, I am afraid that using the same name will cause some confusion. 🤔

PanosCodes: There is an `examples

PanosCodes: There is an `examples` attribute in `Apipie::MethodDescription`, that is never used, if I recall correctly, in the generator. I think it ok to go with `example`

PanosCodes (changes requested):
Could you maybe add a reference to [readme](https://github.com/Apipie/apipie-rails/#swagger-specific-configuration-parameters) too ?

Issue #925, opened 2024-04-05
"allow_blank" setting is not displayed in api doc 


pekopekopekopayo:
The "allow_blank" setting is not indicated in the API documentation.
but allow_nil is displaied
i want to add it
```
path:_params_plain.html.erb, _params.html.erb
        <%= param[:required] ? t('apipie.required') : t('apipie.optional') %>
        <%= param[:allow_nil] ? ', '+t('apipie.nil_allowed') : '' %>
        <%= param[:allow_blank] ? ', '+t('apipie.blank_allowed') : '' %>
```
Of course, I will also include the language

I want to submit a pull request but I don't know how. :cry:

ekohl:
You already identified most of the files you need to change but you'll also need add `blank_allowed` to `config/locales/en.yml` so the translation function can work.

https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github looks like a decent start on how to submit a pull request.

pekopekopekopayo:
@ekohl 
Thank you!! i completed!
So sorry my english is so bad...

ekohl:
Your English is better than my Korean ;)

PR #926, opened 2024-04-05
"allow_blank" setting is not displayed in api doc

pekopekopekopayo:
#925  

"allow_blank" setting is not displayed in api doc

i added blank allowed message


hann-yun:
That's what i want!

mathieujobin (approved):

PR #927, opened 2024-05-09
Bump ruby versions on test jobs

m-nakamura145:
Use latest ruby versions on test jobs


ekohl:
I have never seen this before in a Gemfile:
https://github.com/Apipie/apipie-rails/blob/f300fcc27c7805bcce54b6e02d147bd15931ab12/Gemfile#L8

Reading https://bundler.io/v2.5/man/gemfile.5.html#RUBY I don't really see the point of this. There is already a declaration for compatible versions here:
https://github.com/Apipie/apipie-rails/blob/f300fcc27c7805bcce54b6e02d147bd15931ab12/apipie-rails.gemspec#L14

So I'd propose to drop that line from the Gemfile.

m-nakamura145:
@ekohl I agree with you. removed [733193d](https://github.com/Apipie/apipie-rails/pull/927/commits/733193d324cbf2a2a90de4b3880a7fd40866e01b)

Gemfile
```diff
@@ -5,7 +5,6 @@ source 'https://rubygems.org'
 gemspec path: '.'
 
 # use ENV vars, with default value as fallback for local setup
-ruby(ENV['RUBY_VERSION'] || '3.2.2')
```
mathieujobin, line 8: I guess for a gem it is not as useful as for an app. but this ensure the app runs on the ruby it is supposed to.

people running in docker its usually not a problem, but systems with multiple ruby, `bundler` will complain if the ruby does not have the right version.

I don't have an objection if you prefer removing it.

mathieujobin (approved):

Gemfile
ekohl: Indeed, for an app that doesn't have a gemspec it can make sense but for a gem I don't see the point.

ekohl (approved):

Gemfile
mathieujobin: in a way, it was adding a guarantee that GHA, indeed, runs the test with the correct ruby version.

I don't remember if this project is one of the first that I moved from multiple Gemfiles to a single. but it might be the reason.

mathieujobin:
@ekohl I've been doing squash merge for a few years now, I think we should stick with squash merge on this projct.

PR #929, opened 2024-05-24
Support for Ruby 3.4.0-preview1

m-nakamura145:
This Pull Request adds support for Ruby 3.4.0-preview1.  Running the current code with Ruby 3.4.0-preview1 results in the following error:

```sh
An error occurred while loading ./spec/lib/validators/array_validator_spec.rb. - Did you mean?
                    rspec ./spec/lib/apipie/validator_spec.rb

Failure/Error: require "action_controller/railtie"

LoadError:
  cannot load such file -- bigdecimal
# ./spec/dummy/config/application.rb:3:in '<top (required)>'
# ./spec/dummy/config/environment.rb:5:in '<top (required)>'
# ./spec/spec_helper.rb:10:in '<top (required)>'
# ./spec/lib/validators/array_validator_spec.rb:1:in '<top (required)>'
```

The reason for this error is that in Rails 7.0.X, bigdecimal is not added to the Gemfile (ref: [Rails PR #49039](https://github.com/rails/rails/pull/49039)). Therefore, for Ruby 3.4 and above, we need to use Rails 7.1.X for testing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,8 @@ source 'https://rubygems.org'
 gemspec path: '.'
 
 # use ENV vars, with default value as fallback for local setup
-gem 'actionpack', "~> #{ENV['RAILS_VERSION'] || '7.0'}.0"
-gem 'activesupport', "~> #{ENV['RAILS_VERSION'] || '7.0'}.0"
+gem 'actionpack', "~> #{ENV['RAILS_VERSION'] || RUBY_VERSION >= '3.4' ? '7.1' : '7.0'}.0"
+gem 'activesupport', "~> #{ENV['RAILS_VERSION'] || RUBY_VERSION >= '3.4' ? '7.1' : '7.0'}.0"
 
 gem 'mime-types' # , '~> 3.0'
 gem 'rails-controller-testing'
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,8 @@ source 'https://rubygems.org'
 gemspec path: '.'
 
 # use ENV vars, with default value as fallback for local setup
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -5,8 +5,8 @@ source 'https://rubygems.org'
 gemspec path: '.'
 
 # use ENV vars, with default value as fallback for local setup
-gem 'actionpack', "~> #{ENV['RAILS_VERSION'] || '7.0'}.0"
-gem 'activesupport', "~> #{ENV['RAILS_VERSION'] || '7.0'}.0"
+gem 'actionpack', "~> #{ENV['RAILS_VERSION'] || RUBY_VERSION >= '3.4' ? '7.1' : '7.0'}.0"
+gem 'activesupport', "~> #{ENV['RAILS_VERSION'] || RUBY_VERSION >= '3.4' ? '7.1' : '7.0'}.0"
```
mathieujobin, line 9: conditional is not necessary here, please just make 7.1 default

mathieujobin: and, to keep the conditional you would need parenthesis...
```
irb(main):005> '7.4' || RUBY_VERSION >= '3.2' ? 'yes' : "no"
(irb):5: warning: string literal in condition
=> "yes"
irb(main):006> '7.4' || (RUBY_VERSION >= '3.2' ? 'yes' : "no")
=> "7.4"
irb(main):007> 
```

mathieujobin:
> The reason for this error is that in Rails 7.0.X, bigdecimal is not added to the Gemfile (ref: https://github.com/rails/rails/pull/49039). Therefore, for Ruby 3.4 and above, we need to use Rails 7.1.X for testing.

I think we should also add `bigdecimal` at least in development dependencies of this gem

PR #930, opened 2024-05-30
configure Layout/MultilineMethodCallIndentation to ensure short indent

mathieujobin:
Alternative to #912 

Opinion ?


ekohl (commented):
Feels like more of the code is using the default style, so I'd lean to choosing that, but I don't have a strong opinion on what's better.

mathieujobin (commented):
10 fixed occurences, about the same in terms count

PR #931, opened 2024-06-01
Removed rubocop-challenger

m-nakamura145:
It seems that rubocop-challenger has been disabled for the past four months. 
https://github.com/Apipie/apipie-rails/actions/workflows/rubocop-challenger.yml

Since the rubocop workflow alone is functioning well, how about removing it?


ekohl:
@mathieujobin added this, so I'd like to get his opinion on this.

mathieujobin:
I am disabling it to reduce noise as it creates many PR and takes time to review.

I also don't want to merge too many changes regularly.

I reenable manually to review suggestions from time to time.

I would like to keep it. We could update the schedule but I don't see a lot of value over manually disabling and reenabling.

Gracias

Issue #933, opened 2024-06-04
Blank validator error message is wrong


pekopekopekopayo:
```
class HomeController < ApplicationController
  skip_before_action :verify_authenticity_token
  api!
  param :test_hash, Hash
  def create
    render status: :ok, json: { message: 'Hello, World!' }
  end
end
```

this is my test code

i expect this message 
`Apipie::ParamInvalid (Invalid parameter 'test_hash' value blank: Must be a Hash):`
but error message was wrong
`Apipie::ParamInvalid (Invalid parameter 'test_hash' value nil: Must be a Hash):`

pekopekopekopayo:
@ekohl I opened the PR again by my mistake ㅠ_ㅠ

PR #934, opened 2024-06-04
fix blank error message when allow_blank is false and value is blank

pekopekopekopayo:
#933 

param blank validator error message was wrong and i will fix


spec/lib/apipie/param_description_spec.rb
```diff
@@ -305,7 +305,7 @@
         let(:validation_value) { '' }
 
         it 'raises an error' do
-          expect { subject }.to raise_error(Apipie::ParamInvalid)
+          expect { subject }.not_to raise_error(Apipie::ParamInvalid)
```
pekopekopekopayo, line 308: i think this test code is wrong or 188row test code is wrong or my bad?

pekopekopekopayo:
@ekohl  sorry can i get one more review for me?

spec/support/custom_bool_validator.rb
```diff
@@ -1,4 +1,8 @@
 class CustomBoolValidator < Apipie::Validator::BaseValidator
+  def valid?(value)
+    validate(value)
+  end
```
mathieujobin, line 4: method alias or do without entirely?

lib/apipie/validator.rb
```diff
@@ -480,6 +490,9 @@ def self.validate(value)
     end
 
     class BooleanValidator < BaseValidator
+      def valid?(value)
+        validate(value)
+      end
```
mathieujobin, line 495: much value putting this in multiple validator ?

spec/lib/apipie/param_description_spec.rb
mathieujobin: if it used to raise an error but no longer, there must be a reason.
if that is the new expectation, we should update the test title

mathieujobin: when using `not_to raise_error` we should remove the exception class.
because a different expectation might be raised instead here, creating a false positive

lib/apipie/param_description.rb
```diff
@@ -120,11 +120,11 @@ def validate(value)
       return true if allow_nil && value.nil?
       return true if allow_blank && value.blank?
       value = normalized_value(value)
-      if (!allow_nil && value.nil?) || (blank_forbidden? && value.blank?) || !validator.valid?(value)
```
mathieujobin, line 123: I don't know if we can remove the `!allow_nil` check and the `blank_forbidden?` check.
I think they were added for a reason...

I don't understand what problem we are trying to fix here.

mathieujobin:
closing for now, reopen if you think there is still value.
I think new tests should be added and existing test should be passing, unless the change makes sense.

sorry if its just me who don't understand.

PR #935, opened 2024-06-04
fix blank error message when allow_blank is false and value is blank

pekopekopekopayo:


lib/apipie/param_description.rb
```diff
@@ -120,11 +120,11 @@ def validate(value)
       return true if allow_nil && value.nil?
       return true if allow_blank && value.blank?
       value = normalized_value(value)
-      if (!allow_nil && value.nil?) || (blank_forbidden? && value.blank?) || !validator.valid?(value)
-        error = validator.error
-        error = ParamError.new(error) unless error.is_a? StandardError
-        raise error
-      end
+      return true if !validator.valid?(value)
```
ekohl, line 123: Shouldn't this be:
```suggestion
      return true if validator.valid?(value)
```

ekohl (commented):
It would be good to have tests for this.

lib/apipie/param_description.rb
pekopekopekopayo: yes it is my miss but test code it bothers me ㅠ_ㅠ

pekopekopekopayo:
sorry i will this pr
becuse this pr is duplicate ㅠ_ㅠ sorry....

PR #938, opened 2024-06-25
Fix error for hash object warnings with delegated method descriptions

h-lame:
Fix how we generate warnings about params or properties defined as Hashes without defining what keys they can have inside them.

Our swagger docs stopped working with an ArgumentError:

    ArgumentError: wrong number of arguments (given 0, expected 1)

And it turns out it was in the `Apipie::Generator::Swagger::ParamDescription::Type#warn_hash_without_internal_typespec` method where we call `@controller_method.method` and `@controller_method` was a `Apipie::Generator::Swagger::MethodDescription::Decorator` instance instead of a `Apipie::MethodDescription` instance.  The `Decorator` is a `SimpleDelegator` onto a `MethodDescription` so it _should_ be able to call `method` with no arguments, but it turns out that `SimpleDelegator` won't delegate `method` because it keeps it to refer to `Object#method`.  In #865 we introduced the `method_name` method on `MethodDescription` to address this very problem, but we didn't actually use it in all the places we needed to. This PR fixes that in `warn_hash_without_internal_typespec` where my code was breaking and in `Apipie::Generator::Swagger::ParamDescription::Builder#warn_optional_without_default_value` where we weren't trying to call `method` but we probably should have been.  The error here was getting `#<Apipie::MethodDescription:0x0000000126316f60>` in log messages when we'd rather get the method name.

See https://github.com/Apipie/apipie-rails/pull/939 for some commits in this PR that should be merged separately - they're about a set of fixes that get us to a green build on modern ruby, rack, rubocop infrastructure.  For now, focus on the "Use `method_name`..." commits in this PR.

### DIFF
--- a/lib/apipie/generator/swagger/param_description/builder.rb
+++ b/lib/apipie/generator/swagger/param_description/builder.rb
@@ -1,7 +1,7 @@
 class Apipie::Generator::Swagger::ParamDescription::Builder
   # @param [Apipie::ParamDescription] param_description
   # @param [TrueClass, FalseClass] in_schema
-  # @param [Apipie::MethodDescription] controller_method
+  # @param [Apipie::MethodDescription, nil] controller_method
   def initialize(param_description, in_schema:, controller_method:)
     @param_description = param_description
     @in_schema = in_schema
@@ -98,7 +98,7 @@ class Apipie::Generator::Swagger::ParamDescription::Builder
   def warn_optional_without_default_value(definition)
     if !required? && !definition.key?(:default)
       method_id =
-        if @param_description.is_a?(Apipie::ResponseDescriptionAdapter::PropDesc)
+        if @controller_method.present? && @param_description.is_a?(Apipie::ResponseDescriptionAdapter::PropDesc)
           @controller_method.method_name
         else
           Apipie::Generator::Swagger::MethodDescription::Decorator.new(@controller_method).ruby_name
--- a/lib/apipie/generator/swagger/param_description/builder.rb
+++ b/lib/apipie/generator/swagger/param_description/builder.rb
@@ -99,7 +99,7 @@ class Apipie::Generator::Swagger::ParamDescription::Builder
     if !required? && !definition.key?(:default)
       method_id =
         if @param_description.is_a?(Apipie::ResponseDescriptionAdapter::PropDesc)
-          @controller_method
+          @controller_method.method_name
         else
           Apipie::Generator::Swagger::MethodDescription::Decorator.new(@controller_method).ruby_name
         end
--- a/lib/apipie/generator/swagger/param_description/type.rb
+++ b/lib/apipie/generator/swagger/param_description/type.rb
@@ -114,7 +114,7 @@ class Apipie::Generator::Swagger::ParamDescription::Type
   def warn_hash_without_internal_typespec
     method_id =
       if @param_description.is_a?(Apipie::ResponseDescriptionAdapter::PropDesc)
-        @controller_method.method
+        @controller_method.method_name
       else
         Apipie::Generator::Swagger::MethodDescription::Decorator.new(@param_description.method_description).ruby_name
       end
--- a/lib/apipie/generator/swagger/param_description/type.rb
+++ b/lib/apipie/generator/swagger/param_description/type.rb
@@ -114,7 +114,11 @@ class Apipie::Generator::Swagger::ParamDescription::Type
   def warn_hash_without_internal_typespec
     method_id =
       if @param_description.is_a?(Apipie::ResponseDescriptionAdapter::PropDesc)
-        @controller_method.method_name
+        if @controller_method.present?
+          @controller_method.method_name
+        else
+          Apipie::Generator::Swagger::MethodDescription::Decorator.new(nil).ruby_name
+        end
       else
         Apipie::Generator::Swagger::MethodDescription::Decorator.new(@param_description.method_description).ruby_name
       end
--- a/spec/lib/apipie/generator/swagger/param_description/builder_spec.rb
+++ b/spec/lib/apipie/generator/swagger/param_description/builder_spec.rb
@@ -145,6 +145,24 @@ describe Apipie::Generator::Swagger::ParamDescription::Builder do
             /is optional but default value is not specified/
           ).to_stderr
         end
+
+        context 'and param is a prop desc with a delegated controller method' do
+          let(:param_description) do
+            Apipie.prop(:param, 'object', param_description_options, [])
+          end
+
+          let(:method_desc) do
+            Apipie::Generator::Swagger::MethodDescription::Decorator.new(
+              Apipie::MethodDescription.new(:show, resource_desc, dsl_data)
+            )
+          end
+
+          it 'warns' do
+            expect { subject }.to output(
+              /is optional but default value is not specified/
+            ).to_stderr
+          end
+        end
       end
     end
   end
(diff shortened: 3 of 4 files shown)


mathieujobin:
@PanosCodes opinion ?

lib/apipie/error_description.rb
```diff
@@ -17,9 +17,13 @@ def initialize(code_or_options, desc = nil, options = {})
...
         @description = code_or_options[:desc] || code_or_options[:description]
       else
-        @code = 
+        @code =
           if code_or_options.is_a? Symbol
-            Rack::Utils::SYMBOL_TO_STATUS_CODE[code_or_options]
+            begin
+              Rack::Utils.status_code(code_or_options)
+            rescue ArgumentError
+              nil
```
mathieujobin, line 25: why `nil` here? instead of `code_or_options` ?

lib/apipie/response_description.rb
```diff
@@ -88,11 +88,16 @@ def initialize(method_description, code, options, scope, block, adapter)
 
       @method_description = method_description
 
-      if code.is_a? Symbol
-        @code = Rack::Utils::SYMBOL_TO_STATUS_CODE[code]
-      else
-        @code = code
-      end
+      @code =
+        if code.is_a? Symbol
```
mathieujobin, line 92: I think `if` blocks support `rescue` like ...

```ruby
if cond
  code
rescue => e
  default
else
  default
end
```

mathieujobin (approved):
if all tests and rubocop passes, I don't have any objections.
but I would like @PanosCodes or someone else to chime in for final approval.

lib/apipie/error_description.rb
h-lame: Preserves the old behaviour - if `code_or_options` was a symbol and that symbol wasn't present in `Rack::Utils::SYMBOL_TO_STATUS_CODE` then the return would be `nil`.  Happy to change to `code_or_options` though if you think that's better - arguably it'd probably be more useful to have `code_or_options` and it's closer to the "not a symbol" code path.

lib/apipie/response_description.rb
h-lame: 🤦 - yes, they almost certainly do!  Will tidy up.

h-lame: Actually, no they don't.  You get 
```ruby
rescue => e
  code
else
  other code
end
```

but not:

```ruby
if cond
  code
rescue => e
  other code
else 
  more code
end
```

h-lame:
> if all tests and rubocop passes, I don't have any objections. but I would like @PanosCodes or someone else to chime in for final approval.

I've addressed the rubocop problems - needed to introduce `rubocop-rspec_rails` and fix an existing offence to get to green - then tidy up my own introduced offence.  Both of those in new commits.

Fixed the specs I broke, but there's one that is failing on master for me about rack middleware - I suspect it's because it's using rack 3 now on a fresh install, so let me know if you want me to tidy it up as part of this PR or not.  I've already tidied up a few things to get to running rspec + rubocop so it definitely fits - but I also don't want to scope creep or introduce loads of unnecessary change.

lib/apipie/response_description.rb
mathieujobin: I just thought that's what rubocop was complaining about without looking at the output, sorry for the bad suggestion.

mathieujobin:
only one issue with rails 7.1
image

you don't need to worry about rails 5.x and ruby 2.6, someone else increased the minimum ruby without updating the build matrix.

lib/apipie/generator/swagger/param_description/builder.rb
```diff
@@ -98,8 +98,8 @@ def required_from_path?
   def warn_optional_without_default_value(definition)
     if !required? && !definition.key?(:default)
       method_id =
-        if @param_description.is_a?(Apipie::ResponseDescriptionAdapter::PropDesc)
-          @controller_method
+        if @controller_method.present? && @param_description.is_a?(Apipie::ResponseDescriptionAdapter::PropDesc)
```
PanosCodes, line 101: Can we end up with a blank `@controller_method` ? 
If so maybe we could add a test case for that, update the builder's YARD block to

```ruby
# @param [Apipie::MethodDescription, nil] controller_method
```

PanosCodes (changes requested):
Nice , thank you for your time 🙏

It's not clear to me why we need rubocop-rspec_rails in this PR, could we maybe have it in a separate one ? 🤷

h-lame:
> It's not clear to me why we need rubocop-rspec_rails in this PR, could we maybe have it in a separate one ? 🤷

The commits with `rubocop-rspec_rails` and the `rack` changes are all in service of getting to a green build from a fresh checkout of the repo.  I agree they're making the actual change I want to make less clear.  I'll separate them out into a "get the build green" PR and "fix the warnings about hashes" PR.

lib/apipie/generator/swagger/param_description/builder.rb
h-lame: I admit that I haven't _really_ dug into the "could `@controller_method` be `nil`" question.  I added the `@controller_method.present?` check because of a test failing.  I don't _really_ know if in the normal usage of these objects to define API docs you would ever have a `nil` in there.  But, obviously, the code doesn't actually prevent it.

I could take the check out and try to trace the test that failed a bit deeper to understand why it's able to construct the world where it's exercising these objects without a `@controller_method` being set.  I don't really know what else might break elsewhere if it's `nil`.

h-lame:
> you don't need to worry about rails 5.x and ruby 2.6, someone else increased the minimum ruby without updating the build matrix.

Good to know, although it looks like gemspec still says ruby >= 2.6 and rails > 5?  Should I update those references?

h-lame:
> I'll separate them out into a "get the build green" PR and "fix the warnings about hashes" PR.

I've extracted those commits out to https://github.com/Apipie/apipie-rails/pull/939 - and also fixed the broken spec about the recorder middleware caused by using rack 3.

If we can merge that then I can rebase this against that and we can focus on the two "Use `method_name` ..." commits that are the actual change we care about for this PR.

h-lame:
Rebased now that #939 is merged and that makes the change much smaller and contained.  To address is the outstanding comment on a blank `@controller_method` - I'll try to trace that through and see if it's the test being setup weird, or an expected way of using the code.

mathieujobin:
> I'll try to trace that through and see if it's the test being setup weird, or an expected way of using the code.

@h-lame if you could add a test that covers/prove its well handled that'd be most perfect.

mathieujobin (approved):
Looking at this one more time, LGTM ! I think I will merge next week unless objections

lib/apipie/generator/swagger/param_description/builder.rb
h-lame: A `nil` `@controller_method` is expected if we call `json_schema_for_self_describing_class` (see https://github.com/Apipie/apipie-rails/blob/4be37413e447a39dd7c117c5c6fd8fad41d3ddf9/lib/apipie/swagger_generator.rb#L36-L43).  I'll update the YARD block and introduce an explicit test for it being `nil`.

h-lame: Ok, added a new commit that changes the YARD for `Apipie::Generator::Swagger::ParamDescription::Builder` and puts in tests for both Apipie::Generator::Swagger::ParamDescription::Builder and Apipie::Generator::Swagger::ParamDescription::Type when emitting warnings with `PropDesc` for a `param_description` and `nil` for a `controller_method`.

mathieujobin:
quick compare link for the last force-push
https://github.com/Apipie/apipie-rails/compare/28ba4a7..6d84540

Thanks

PR #939, opened 2024-07-05
Get the build green for modern ruby and rack and rubocop-rspec

h-lame:
Separated out from #938 this PR attempts to get the master build to green from a fresh checkout.

These days you're probably on ruby 3.x and will get rack 3.x and rubocop-rspec 3.x too.  These new versions all cause problems with the code in master, so the build won't pass.

1.  https://github.com/rack/rack/pull/2137 - Handle the fact that the values in `Rack::Utils::SYMBOL_TO_STATUS_CODE` are now different and this causes the dummy rails app we use for testing to not load and this blocks the entire suite from running.  The solution is to use `Rack::Utils.status_code` which has a workaround for the old names.
2. https://github.com/rack/rack/pull/2018 - Handle the fact that `rack.input` is now optional in the rack env. The solution is to use safe navigation to call the methods we wanted.
3. https://github.com/rubocop/rubocop-rspec/pull/1848 - Introduce `rubocop-rspec_rails` as the cops it provides were extracted from `rubocop-rspec` and our config in `rubocop_todo.yml` referenced them and rubocop doesn't like config for cops it doesn't know about so that blocked the build.
4. Fix the one rubocop offence that being able to run the build again let us see.

### DIFF
--- a/apipie-rails.gemspec
+++ b/apipie-rails.gemspec
@@ -38,6 +38,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rubocop-rails'
   s.add_development_dependency 'rubocop-rspec'
   s.add_development_dependency 'rubocop-performance'
+  s.add_development_dependency 'rubocop-rspec_rails'
   s.add_development_dependency "simplecov"
   s.add_development_dependency "sqlite3"
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,7 @@
 require:
   - rubocop-rails
   - rubocop-rspec
+  - rubocop-rspec_rails
   - rubocop-performance
 
 inherit_from: .rubocop_todo.yml
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -1155,7 +1155,7 @@ RSpec/PredicateMatcher:
 # Offense count: 1
 # This cop supports unsafe autocorrection (--autocorrect-all).
 # Configuration parameters: Inferences.
-RSpec/Rails/InferredSpecType:
+RSpecRails/InferredSpecType:
   Exclude:
     - 'spec/controllers/pets_controller_spec.rb'
 
--- a/lib/apipie/error_description.rb
+++ b/lib/apipie/error_description.rb
@@ -17,9 +17,13 @@ module Apipie
         @metadata = code_or_options[:meta]
         @description = code_or_options[:desc] || code_or_options[:description]
       else
-        @code = 
+        @code =
           if code_or_options.is_a? Symbol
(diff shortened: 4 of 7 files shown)


mathieujobin (approved):
👍🏽 LGTM

apipie-rails.gemspec
```diff
@@ -38,6 +38,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rubocop-rails'
   s.add_development_dependency 'rubocop-rspec'
   s.add_development_dependency 'rubocop-performance'
+  s.add_development_dependency 'rubocop-rspec_rails'
```
mathieujobin, line 41: can you try adding ` if RUBY_VERSION >= "2.7"` please?

or take out this line for now, ... if you can

h-lame: I can do that, but the problem will probably then be that ruby 2.6 rubocop will complain about `RSpecRails/InferredSpecType` being a cop it doesn't understand whereas ruby 2.7+ rubocop complains about `RSpec/Rails/InferredSpecType`.  Can we put conditional rubocop_todo by ruby version?

In another thread I think you suggested someone had dropped ruby 2.6 / rails 5 support - should we make that official?

h-lame: I'll push up a commit with the `if` version comparison so we can test that theory though

mathieujobin:
@h-lame The condition is cool, we just need to add a rubocop exclusion for it, I think its fine, I will probably remove support for ruby 2.6 soon anyway, not much of a point anymore. bit in the mean time. its good.

h-lame:
Thanks!  Reading up on the now failing rubocop complaint about using RUBY_VERSION in the gem spec, sounds like the preferred option is moving all the `add_development_dependency` things to Gemfile (see https://github.com/rubygems/rubygems/issues/7799#issuecomment-2192720316) - any interest in me doing that?

mathieujobin:
@h-lame No thanks, the condition is for ruby 2.6, which we'll probably remove support for. so that exclusion will go away anyway soon.

about the comment you linked on rubygems, I don't see how rubygems will be able to deprecate `add_development_dependency`, I guess they can all be moved to the Gemfile instead but ... I think a lot of gem owners will not be happy about that.

thanks for your contribution to Apipie

h-lame:
> about the comment you linked on rubygems, I don't see how rubygems will be able to deprecate `add_development_dependency`, I guess they can all be moved to the Gemfile instead but ... I think a lot of gem owners will not be happy about that.

Yeah, I agree, seems like a pretty big breaking change to leave in a comment on an unrelated issue.

> thanks for your contribution to Apipie

No worries!  Thanks for accepting it.

PR #944, opened 2024-10-21
Support non-string enum validators ...

jaynetics:
... both for custom validators and self-describing properties.

Before:

```ruby
expect(pm_schema).to have_field(:num_tails, 'number', {:description => "Number of tails", :enum => [0, 1]})
# => expected param 'num_tails' to have type 'number' (got 'string')

expect_param_def("get", "/users/by_department_number", "department", "type", "integer")
# =>        expected: "integer"
# =>            got: "string"
```

After:

tests pass.

This PR also adds support for subclassing of validators.


lib/apipie/validator.rb
```diff
@@ -25,8 +25,11 @@ def inspect
       end
 
       def self.inherited(subclass)
+        BaseValidator.validators.unshift(subclass)
```
jaynetics, line 28: Previously when subclassing a concrete validator, e.g. `EnumValidator`, this method would init the `@validators` ivar on `EnumValidator` and insert the new subclass there.

By explicitly referencing `BaseValidator`, the new class is added to the correct list.

This change shouldn't affect backwards compatibility. Worst case if anyone subclassed existing validators, they probably did `BaseValidator.inherited(my_validator)` or so, in which case this fix would cause `my_validator` to appear in the list twice.

spec/lib/swagger/rake_swagger_spec.rb
```diff
@@ -135,12 +138,15 @@ def expect_response_params_def(http_method, path, response_code, param_name, fie
...
+        expect_param_def("get", "/users/by_department_number", "department", "type", "integer")
+        expect_param_def("get", "/users/by_department_number", "department", "enum", [1, 2, 3])
+
         expect_param_def("get", "/users/in_departments", "departments", "in", "query")
         expect_array_param_def("get", "/users/in_departments", "departments",
                          %w[finance operations sales marketing HR])
 
         expect_tags_def("get", "/twitter_example/{id}/followers", %w[twitter_example following index search])
-
+        expect_response_params_def("get", "/pets/{id}/as_properties", 200, "pet_name", "example", "mypet")
```
jaynetics, line 149: this added line is not related to this PR. i've just copied it from the other changed example above because it looks as if the two examples were supposed to mirror each other.

PR #946, opened 2024-11-05
Support passing a callable to EnumValidator

ekohl:
In case the possible values are dynamic. This probably can't ever be made to work reliably with statically generating HTML documentation.

It's expected that the proc never throws an exception.


ekohl:
When writing this I didn't realize it would become ambiguous when to use `ProcValidator` and `EnumValidator`.

mathieujobin (commented):
if you can increase test coverage and make sure all tests pass. then I can merge and release.

@PanosCodes if you have time to review later, that'd be appreciated as well.

lib/apipie/validator.rb
```diff
@@ -177,23 +177,23 @@ def description
...
       end
 
       def validate(value)
-        @array.include?(value)
+        values.include?(value)
       end
 
       def self.build(param_description, argument, options, proc)
-        self.new(param_description, argument) if argument.is_a?(Array)
+        self.new(param_description, argument) if argument.is_a?(Array) || argument.respond_to?(:call)
```
ekohl, line 188: @mathieujobin I'm not sure how to disambiguate this from the `ProcValidator` here:
https://github.com/Apipie/apipie-rails/blob/eeef4de93ac7fe68942cb0e878a9e8353eceeb17/lib/apipie/validator.rb#L313

If you have any thoughts, I'd love to hear them.

I thought about introducing a class that somehow wraps the callable, but it feels rather heavy.

mathieujobin: Because EnumValidator will receive a Proc as well ? 
Can you check that is it an Enum before? 

`.respond_to?(:call)` can be a lot of things... so, you accept to build a Enum validator with either an Array or a Proc? not an Enum ? is a Enum a Proc underneath?

it would be good to add a new test where the confusion between the two validator can be demonstrated and the right one is being selected. this will help finding a working implementation.

PanosCodes:
I don't know if it's possible, but since we have a dynamic argument that we want to evaluate at the time apipie is called, can we maybe get the value of the callable before it is given to the validators and have the validator work as is? 🤔

PR #948, opened 2024-12-19
Rails 72 fix deprecation warn

Migoo:
Fix deprecation method ActiveSupport::Deprecation.warn. Replaced with ActiveSupport::Deprecation.new.warn

### DIFF
--- a/lib/apipie/application.rb
+++ b/lib/apipie/application.rb
@@ -380,7 +380,7 @@ module Apipie
 
     # @deprecated Use {#get_resource_id} instead
     def get_resource_name(klass)
-      ActiveSupport::Deprecation.new.warn(
+      ActiveSupport::Deprecation.new('2.0', 'apipie-rails').warn(
         <<~HEREDOC
           Apipie::Application.get_resource_name is deprecated.
           Use `Apipie::Application.get_resource_id instead.
--- a/lib/apipie/application.rb
+++ b/lib/apipie/application.rb
@@ -380,7 +380,7 @@ module Apipie
 
     # @deprecated Use {#get_resource_id} instead
     def get_resource_name(klass)
-      ActiveSupport::Deprecation.warn(
+      ActiveSupport::Deprecation.new.warn(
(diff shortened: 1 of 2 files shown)


lib/apipie/application.rb
```diff
@@ -380,7 +380,7 @@ def active_dsl?
 
     # @deprecated Use {#get_resource_id} instead
     def get_resource_name(klass)
-      ActiveSupport::Deprecation.warn(
+      ActiveSupport::Deprecation.new.warn(
```
ekohl, line 383: Reading https://api.rubyonrails.org/classes/ActiveSupport/Deprecation.html I think you should really set a specific deprecation horizon. So something like:
```suggestion
      ActiveSupport::Deprecation.new('2.0', 'apipie-rails').warn(
```

ekohl (changes requested):

mathieujobin (changes requested):
I agree with @ekohl 's comment.

ekohl:
@Migoo the comment was applicable to to all lines changed and I just didn't want to repeat myself. Could you please apply it everywhere consistently?

Migoo:
> @Migoo the comment was applicable to to all lines changed and I just didn't want to repeat myself. Could you please apply it everywhere consistently?

My bad ! I just push for all deprecation

ekohl (approved):
Code looks good to me so I just approved the tests to see if CI agrees.

@mathieujobin do you agree this is now good to go? I also think we should squash the commits on merge.

ekohl:
Looking at the tests, I see we're not testing on Rails 7.2 so perhaps it should also be added:
https://github.com/Apipie/apipie-rails/blob/62b24b6a2815613f1f9575dc218bdaf425a16988/.github/workflows/build.yml#L11

We can also ignore the Ruby 3.4 failure in this PR, but we should address that as well as change it to Ruby 3.4 final now that's out.

evgeni (approved):

PR #951, opened 2025-02-19
Fix issue with loading Rails logger

ehelms:
See https://stackoverflow.com/questions/79360526/uninitialized-constant-activesupportloggerthreadsafelevellogger-nameerror

### DIFF
--- a/spec/dummy/config/boot.rb
+++ b/spec/dummy/config/boot.rb
@@ -9,4 +9,4 @@ end
 
 $:.unshift File.expand_path('../../../lib', __dir__)
 
-require "logger" # Fix concurrent-ruby removing logger dependency which Rails itself does not have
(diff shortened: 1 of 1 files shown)


spec/dummy/config/boot.rb
```diff
@@ -7,4 +7,6 @@
   Bundler.setup
 end
 
-$:.unshift File.expand_path('../../../lib', __dir__)
\ No newline at end of file
+$:.unshift File.expand_path('../../../lib', __dir__)
+
+require "logger" # Fix concurrent-ruby removing logger dependency which Rails itself does not have
```
ekohl, line 12: While there are more RuboCop violations, let's not introduce a new one :)
```suggestion
require 'logger' # Fix concurrent-ruby removing logger dependency which Rails itself does not have
```

ekohl (changes requested):

ekohl (approved):

PR #952, opened 2025-04-23
update jQuery to 3.7.1

markbluhm:
Uses [slim build](https://code.jquery.com/jquery-3.7.1.slim.min.js) because Bootstrap says [that's what it uses](https://getbootstrap.com/docs/4.3/getting-started/introduction/#js).
> We use [jQuery’s slim build](https://blog.jquery.com/2016/06/09/jquery-3-0-final-released/), but the full version is also supported.

This is to address CVEs that may or may not matter but are often caught in security scans because the asset included is from a vulnerable jQuery version.

I didn't update bootstrap or bootstrap-collapse. The app we have that uses this only uses bootstrap-collapse so I didn't test any of the other components but collapse is still working fine.


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
  Changelog
 ===========
 
+## [v1.4.3](https://github.com/markbluhm/apipie-rails/tree/v1.4.2) (2025-04-22)
```
AidanHibbard, line 4: Is this meant to be pointed at your fork still?

markbluhm: fixed ty

aidanhibbard (approved):

PR #957, opened 2025-07-15
Replace Maruku with Commonmarker for Markdown

shqear93:


ekohl:
Last update in 2017, that's old.

I'm looking at commonmarker and I see it's pulling in a rust crate. That does significantky increase toolchain size. Not sure if we should mandate that. Also requires Ruby 3.2 and in our current version we still carry older support.

I kicked off CI to see how that behaves

shqear93:
@ekohl what do you think about redcarpet? https://github.com/vmg/redcarpet

mathieujobin:
I think pulling rust is acceptable. its should only be pulled once per ruby version, and more and more gems will be using it.

mathieujobin (commented):
Can you update the changelog with your entry? Respecting current style?

evgeni:
I have no idea about Rust, but I know @odilhao builds Rust things for Python already

Odilhao:
> I have no idea about Rust, but I know @Odilhao builds Rust things for Python already

Rust bindings requires me to always vendor the [crates](https://downloads.theforeman.org/vendor/) every update, it's a really fun process.

PR #962, opened 2025-10-27
Fix Rails 8.1 router keyword args deprecation warning

jamesds:
Rails 8.1 is raising a deprecation warning on a single line in lib/apipie/routing.rb, where `get` (and all other router methods) will be expected to use keyword arguments from Rails 8.2 onwards.

The initial fix failed in the CI tests for the oldest 2 ruby/rails versions, so I've amended the fix to only apply to Rails 5.2 and above, if the older versions still need to be supported.

### DIFF
--- a/lib/apipie/routing.rb
+++ b/lib/apipie/routing.rb
@@ -5,7 +5,7 @@ module Apipie
         namespace "apipie", :path => Apipie.configuration.doc_base_url do
           get 'apipie_checksum', :to => "apipies#apipie_checksum", :format => "json"
           constraints(:version => %r{[^/]+}, :resource => %r{[^/]+}, :method => %r{[^/]+}) do
-            get(options.reverse_merge("(:version)/(:resource)/(:method)" => "apipies#index", :as => :apipie))
+            get(**options.reverse_merge("(:version)/(:resource)/(:method)" => "apipies#index", :as => :apipie))
           end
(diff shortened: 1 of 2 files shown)


mathieujobin:
@copilot can you check whats wrong with Rubocop ?

lib/apipie/routing.rb
```diff
@@ -5,7 +5,9 @@ def apipie(options = {})
         namespace "apipie", :path => Apipie.configuration.doc_base_url do
           get 'apipie_checksum', :to => "apipies#apipie_checksum", :format => "json"
           constraints(:version => %r{[^/]+}, :resource => %r{[^/]+}, :method => %r{[^/]+}) do
-            get(options.reverse_merge("(:version)/(:resource)/(:method)" => "apipies#index", :as => :apipie))
+            get_args = options.reverse_merge("(:version)/(:resource)/(:method)" => "apipies#index", :as => :apipie)
+
+            Rails.version >= "5.2" ? get(**get_args) : get(get_args)
```
mathieujobin, line 10: lets remove support for Rails 5.x and 6.x @copilot
and update CHANGELOG and bump the next minor version

mathieujobin:
Thanks for working on this @jamesds

jamesds:
Thank you @mathieujobin!

PR #965, opened 2025-11-28
Add support for x-extension properties in Apipie.prop

mscoutermarsh:
In our API spec, we found the need recently to set [`x-extension`](https://swagger.io/docs/specification/v2_0/swagger-extensions/) properties for some of our downstream API spec consumers (which auto generate SDKs based on our api spec).

This PR adds support for setting `x-extensions`.

## Example:
```ruby
Apipie.prop(:password, "password", { description: "The plain text password", "x-param-sensitive": true })
```

```json
"password": {
  "type": "string",
  "description": "The plain text password, available only after create",
  "x-param-sensitive": true
}
```


no-itsbackpack (approved):

PanosCodes (approved):
Nice

PR #972, opened 2026-07-19
Allow overriding the generated operationId per route

bogdan:
## Summary

Right now the swagger `operationId` for every route is always derived purely from the HTTP method + path (`Apipie::Generator::Swagger::OperationId`), e.g. `get_api_users_id`. There's no way to override it, which is awkward for consumers of the generated OpenAPI spec (e.g. client code generators) that key off `operationId` for method naming.

This adds an `:operation_id` option to the `api` DSL, following the exact same pattern already used by the existing `:deprecated` option (read from `api.options`):

```ruby
api :GET, "/api/users/:id", "Show user", :operation_id => "showUser"
```

If `:operation_id` isn't given, behavior is unchanged (falls back to the existing auto-derived id).

## Changes

- `lib/apipie/generator/swagger/method_description/api_schema_service.rb`: prefer `api.options[:operation_id]` over the auto-generated id when present.
- `spec/.../api_schema_service_spec.rb`: added coverage for both the default and custom-id cases.
- `README.md`: documented the new option alongside the other swagger-specific DSL options.

## Test plan

- [x] `bundle exec rspec spec/lib/apipie/generator/swagger/method_description/api_schema_service_spec.rb spec/lib/apipie/generator/swagger/operation_id_spec.rb` — 19 examples, 0 failures
- [x] Verified no behavior change when `:operation_id` is omitted (existing tests for auto-derived ids still pass)

### DIFF
--- a/lib/apipie/generator/swagger/method_description/api_schema_service.rb
+++ b/lib/apipie/generator/swagger/method_description/api_schema_service.rb
@@ -10,7 +10,7 @@ class Apipie::Generator::Swagger::MethodDescription::ApiSchemaService
     @method_description.apis.each_with_object({}) do |api, paths|
       api = Apipie::Generator::Swagger::MethodDescription::ApiDecorator.new(api)
       path = Apipie::Generator::Swagger::PathDecorator.new(api.path)
-      op_id = Apipie::Generator::Swagger::OperationId.from(api).to_s
+      op_id = api.options[:operation_id] || Apipie::Generator::Swagger::OperationId.from(api).to_s
 
       if Apipie.configuration.generator.swagger.generate_x_computed_id_field?
         Apipie::Generator::Swagger::ComputedInterfaceId.instance.add!(op_id)
--- a/README.md
+++ b/README.md
@@ -1748,6 +1748,18 @@ example:
 param :do_something, Boolean, :desc => "take an action", :required => false, :default_value => false
 ```
 
+## Specifying a custom operationId
+
+By default, the `operationId` of each generated swagger operation is
+derived from its HTTP method and path (e.g. `get_api_users_id`). To
+use a custom value instead (for example to match the naming your
+client code generator expects), pass `:operation_id` in the `api` DSL
+call:
+
+``` ruby
+api :GET, "/api/users/:id", "Show user", :operation_id => "showUser"
+```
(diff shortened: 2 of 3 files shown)


lib/apipie/generator/swagger/method_description/api_schema_service.rb
```diff
@@ -10,7 +10,7 @@ def call
     @method_description.apis.each_with_object({}) do |api, paths|
       api = Apipie::Generator::Swagger::MethodDescription::ApiDecorator.new(api)
       path = Apipie::Generator::Swagger::PathDecorator.new(api.path)
-      op_id = Apipie::Generator::Swagger::OperationId.from(api).to_s
+      op_id = api.options[:operation_id] || Apipie::Generator::Swagger::OperationId.from(api).to_s
```
ofedoren: This has several issues:
 - api.options[:operation_id] is being used as is, whereas the generated one is ensured to be a string
 - if api.options[:operation_id] is an empty string, `||` would consider it as truthly value, consider using `api.options[:operation_id].to_s.presence`: `to_s` here will ensure it's a string and `presence` would return falsy value if the string is empty, thus the value will be generated instead
  - The generated one should be uniq, but do we ensure that the provided value is uniq as well or leave it to the users?

And as a side note, should we validate the value via `^[a-zA-Z0-9\.\-_]+$`?

ofedoren (commented):
Thanks, @bogdan ! Few notes inline:

lib/apipie/generator/swagger/method_description/api_schema_service.rb
bogdan: Doc says this:

> Unique string used to identify the operation. The id MUST be unique among all operations described in the API. The operationId value is case-sensitive. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.

https://swagger.io/specification/v3.2/

So, validation is `RECOMMENDED` only.

I think uniqueness should be validates accross the board including the generated ones, so it seems to be out of scope of this change because the generated one may collide (very unlikely though).

ofedoren (approved):
Thanks, @bogdan !

I think we can get this in as is.

PR #973, opened 2026-07-20
Respect per-field allow_nil for response bodies

bogdan:
## Summary

`param`/`property`'s existing `allow_nil: true` option (`Apipie::ParamDescription#allow_nil`) is only ever consulted for request params. For response bodies, `Composite#to_swagger` only checks the whole-response `Context#allow_null?` flag — which `ResponseService` hardcodes to `false` — so there was no way to mark an individual response field as nullable without making the *entire* response nullable (`config.generator.swagger.allow_null`, which doesn't exist as a per-field or even per-response knob today).

This adds a per-field check: a field can opt into nullability individually via its own `allow_nil: true`, while the response-level default (currently always `false`) continues to apply to everything else — fully backward compatible.

```ruby
returns code: 200 do
  property :id, Integer
  property :translation, String, required: false, allow_nil: true
end
```
produces
```json
{ "translation": { "type": "string", "x-nullable": true } }
```
instead of previously being indistinguishable from a field that's sometimes entirely absent from the response.

## Changes

- `lib/apipie/generator/swagger/param_description/composite.rb`: replaced the two `@context.allow_null?` checks with `allow_null_for?(param_description)`, which is `param_description.allow_nil || @context.allow_null?`.
- `lib/apipie/response_description_adapter.rb`: `PropDesc` (what self-describing classes / `Apipie.prop` produce) didn't expose `allow_nil` at all — added it the same way `required` is already derived, so `Apipie.prop(:name, "string", allow_nil: true)` works identically to the regular DSL's `property ..., allow_nil: true`.
- `lib/apipie/generator/swagger/param_description/type.rb`: nullability is now expressed via the `x-nullable` vendor extension instead of mutating `type` into a 2-element array (`type: ['string', 'null']`). The array form is valid JSON Schema but not valid Swagger 2.0 — the spec requires a Schema Object's `type` to be a single string — and tooling that expects spec-conformant OAS2 input (verified against `@hey-api/openapi-ts`'s Swagger 2.0 importer) does plain string-equality checks against `type`, so an array value silently fails to match anything and the field degrades to an untyped/`unknown` result instead of picking up the nullability. `x-nullable` is the widely-supported OAS2 convention for this (used by swagger-codegen, NSwag, etc.) and doesn't touch `type` at all. This also incidentally fixes a latent double-wrap in the array + nullable combination, where the same array mutation was applied twice (once for the array wrapper, once again unconditionally afterward); setting a flag twice is a no-op instead.

## Test plan

- [x] New specs: per-field opt-in on a regular `Apipie::ParamDescription`, opt-in while the response-level default stays `false`, the self-describing-class/`PropDesc` path (`Apipie.prop(..., allow_nil: true)`), and the `x-nullable` representation itself (`composite_spec.rb`, `response_schema_service_spec.rb`).
- [x] Verified end-to-end against a real downstream Rails app: generated a Swagger 2.0 doc with a nullable response field, ran it through `@hey-api/openapi-ts` before and after this change — before: `unknown`; after: the expected `T | null`.
- [x] `bundle exec rspec` (full suite): no regressions to the existing whole-response `allow_null: true` behavior.

### DIFF
--- a/lib/apipie/generator/swagger/param_description/composite.rb
+++ b/lib/apipie/generator/swagger/param_description/composite.rb
@@ -48,7 +48,7 @@ class Apipie::Generator::Swagger::ParamDescription::Composite
           schema = for_array(schema)
         end
 
-        if @context.allow_null?
+        if allow_null_for?(param_description)
           schema = with_null(schema)
         end
 
@@ -63,7 +63,7 @@ class Apipie::Generator::Swagger::ParamDescription::Composite
             controller_method: @context.controller_method
           ).
           with_description(language: @context.language).
-          with_type(with_null: @context.allow_null?).
+          with_type(with_null: allow_null_for?(param_description)).
           with_in(
             default_in_value: @context.default_in_value,
             http_method: @context.http_method
@@ -115,5 +115,12 @@ class Apipie::Generator::Swagger::ParamDescription::Composite
     if !param_description.respond_to?(:required)
       raise "Unexpected param_desc format"
     end
+  end
+
+  # A field can opt into being nullable individually (param/property's
+  # existing allow_nil: true option, now also read for response bodies)
+  # even when the surrounding response/context isn't nullable by default.
+  def allow_null_for?(param_description)
+    (param_description.respond_to?(:allow_nil) && param_description.allow_nil) || @context.allow_null?
   end
 end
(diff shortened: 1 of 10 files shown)


ofedoren:
This needs a rebase now

ofedoren:
The failures will be fixed in https://github.com/Apipie/apipie-rails/pull/975.

